PR #16, opened 2020-06-10
Fix EPTS serialization issues with period setting and players not pre…

bdagnino:
…sent on a DataFormatSpecification

### DIFF
--- a/kloppy/infra/serializers/tracking/epts/reader.py
+++ b/kloppy/infra/serializers/tracking/epts/reader.py
@@ -79,9 +79,10 @@ def read_raw_data(raw_data: Readable, meta_data: EPTSMetaData,
             row['timestamp'] = timestamp
 
             row['period_id'] = None
-            for i, p in enumerate(periods):
-                if timestamp >= p.start_timestamp and timestamp <= p.end_timestamp:
-                    row['period_id'] = i+1
+            for period in periods:
+                if period.start_timestamp <= timestamp <= period.end_timestamp:
+                    row['period_id'] = period.id
+                    break
 
             yield row
 
--- a/kloppy/infra/serializers/tracking/epts/reader.py
+++ b/kloppy/infra/serializers/tracking/epts/reader.py
@@ -58,7 +58,6 @@ def read_raw_data(raw_data: Readable, meta_data: EPTSMetaData,
     _set_current_data_spec(0)
 
     periods = meta_data.periods
-    period_idx = 0
     n = 0
     sample = 1. / sample_rate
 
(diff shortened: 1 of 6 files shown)


kloppy/infra/serializers/tracking/epts/serializer.py
```diff
@@ -41,15 +41,27 @@ def _frame_from_row(row: dict, meta_data: EPTSMetaData) -> Frame:
...
-                home_team_player_positions[player.jersey_no] = Point(
-                    x=row[f'player_home_{player.jersey_no}_x'],
-                    y=row[f'player_home_{player.jersey_no}_y']
-                )
+                if f'player_home_{player.jersey_no}_x' in row:
+                    home_team_player_positions[player.jersey_no] = Point(
+                        x=row[f'player_home_{player.jersey_no}_x'],
+                        y=row[f'player_home_{player.jersey_no}_y']
+                    )
+                else:
```
koenvo, line 49: Good one. I would prefer not storing the data at all when info is not there. So just skip the else part. Could you also add a test case with a small file that would break previous implementation?

kloppy/infra/serializers/tracking/epts/serializer.py
```diff
@@ -31,7 +31,7 @@ def __validate_inputs(inputs: Dict[str, Readable]):
     @staticmethod
     def _frame_from_row(row: dict, meta_data: EPTSMetaData) -> Frame:
         timestamp = row['timestamp']
-        if meta_data.periods:
+        if meta_data.periods and row['period_id'] != None:
```
koenvo, line 34: Compare with None should be `row['period_id'] is not None` or if period_id cannot be `0` it can be: `row['period_id']`

kloppy/infra/serializers/tracking/epts/reader.py
```diff
@@ -79,10 +78,10 @@ def _set_current_data_spec(idx):
...
             row['timestamp'] = timestamp
 
-            if period_idx > len(periods):
-                if timestamp > periods[period_idx].end_timestamp:
-                    period_idx += 1
-                row['period_id'] = periods[period_idx].id
+            row['period_id'] = None
+            for i, p in enumerate(periods):
+                if timestamp >= p.start_timestamp and timestamp <= p.end_timestamp:
+                    row['period_id'] = i+1
```
koenvo, line 84: Good catch! Not sure with what file I tested the previous code but that's not working.

I think it's good to use `period.id` instead of `i + 1` to be sure it won't break when period_id's change some day.

```python
for period in periods:
   if period.start_timestamp <= timestamp <= period.end_timestamp:
     row['period_id'] = period.id
     break
```

PR #28, opened 2020-06-26
minor correction to readme

rjtavares:
Fixed the Quickstart section example (transform function with 'to_pitch_dimensions' argument instead of 'pitch_dimensions')


koenvo:
Good catch!

PR #31, opened 2020-07-01
added option to parse additional columns when calling to_pandas

rjtavares:
To test the feature with events:

```
dataset = datasets.load("statsbomb")
get_first_related_event = lambda x: x.raw_event['related_events'][0] if 'related_events' in x.raw_event else None
to_pandas(dataset, additional_columns={'related_event': get_first_related_event, 'match': 'test1'})

```

With tracking data:

```
dataset = datasets.load("metrica_tracking", options={"sample_rate": 1.0 / 12, "limit": 10})
to_pandas(dataset, additional_columns={'attacking_direction': lambda x: x.period.attacking_direction.value, 'match': 'game1'})

```

Do you want me to add a section to the readme explaining how it works?

### DIFF
--- a/README.md
+++ b/README.md
@@ -247,6 +247,21 @@ new_dataset = Transformer.transform_dataset(
 )
 ```
 
+### Adding additional columns to DataFrame
+
+You can add more information to the DataFrame returned by `to_pandas` by passing a Dictionary in the format `{'column_name': function or scalar}`. Function should take a frame (for tracking data) or an event (for event data) as argument and return 
+
+```
+from kloppy import datasets, to_pandas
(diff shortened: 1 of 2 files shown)


koenvo:
Please add a section to the readme. Thanks

One minor code feedback:
can you make `generic_record_converter` a nested function within the `to_pandas` function? Right before `return pd.DataFrame.from_records(map(converter, dataset.records))`. Then you don't need the lambda and it can use `_record_converter` and `additional_columns` from the local scope instead of passing them as arguments.

rjtavares:
Done.

koenvo:
Thank you!

Issue #3, opened 2020-05-08
Metadata model


bdagnino:
Would be useful to have a game model with pre defined fields that get filled automatically if the data is available on the data package, or could get filled in manually. 

Could be based on the fields/inputs present in the FIFA EPTS standard format:
1. Global
2. Teams 
3. Players

This way independently of the data provider you could have the same fields and know whether they are available or not.

koenvo:
While playing around with the pattern matching it's very useful to have metadata to be able to directly see names instead of only jersey numbers. Let's spend some time on this one soon

koenvo:
@bdagnino does it make sense for your usecase to attach a metadata attribute to the dataset containing the metadata?

bdagnino:
@koenvo yes. Let me know if you want any help with the implementation or need any extra feedback.

koenvo:
Would be good to discuss the models in more detail. 

It should include at least:

```python

class Metadata:
   home_team: Team  # this might break for EPTS without teams..
   away_team: Team

class Team:
    team_id: str   # provider ID. This can also be 'home' / 'away' in case provider doesn't include ids
    name: str
    players: List[Player]

class Player:
    player_id: str  # provider ID. must be unique over all players. In case a provider does have non-unique values it must be prefixed with 'home_' or 'away_'. This is the case for TRACAB/Metrica CSV which only has jersey numbers
    jersey_number: int
    position: Position
    name: str
    first_name: str
    last_name: str

class Position:
   position_id: str
   name: str
```

bdagnino:
Nice! Solid base! Additions I think are important:

1. Pitch for the dimensions of the field, it would be the same Pitch model already available the models/common.py
2. frame_rate
3. score
4. Periods, same as in models/common.py

In the future we could consider adding Subtitutions, but not necesary for now I believe. 

So the model would be:

```python
class Metadata:
   home_team: Team  # this might break for EPTS without teams..
   away_team: Team
   pitch: Pitch
   score: List # first home, second away [0,0]
   periods: List[Period]
   frame_rate: float


class Team:
    team_id: str   # provider ID. This can also be 'home' / 'away' in case provider doesn't include ids
    name: str
    players: List[Player]

class Player:
    player_id: str  # provider ID. must be unique over all players. In case a provider does have non-unique values it must be prefixed with 'home_' or 'away_'. This is the case for TRACAB/Metrica CSV which only has jersey numbers
    jersey_number: int
    position: Position
    name: str
    first_name: str
[...]

PR #32, opened 2020-07-04
Bruno/add dataset metadata

bdagnino:
In this pull request I have modified the Player, Team, MetaData dataclasses as discussed in #3. Having the metadata information as part of the dataset, now the pandas dataframe generated when calling `to_pandas(dataset)` has the player IDs provided by the data provider as columns names. 

The idea of this pull request is not to merge it, but to evaluate whether the approach taken and the modifications made are ok. If so, I'll extend the same changes to the other formats (tracab, metrica, etc).

Some questions/comments on my side are:

1. I changed the Team dataclass to be a standard dataclass (not an Enum). On what I did I don't see a reason to keep the Enum as we can base it on the team ID or "home" / "away" as IDs when no IDs are provided by the provider. However as I'm not familiar with all the code base I'm not sure whether we prefer to keep it that way. 
2. At the moment the column names are are `{players_id}_{channel_id}`. I didn't add a `player_` at the star because I thought it was redundant, but when checking the tests for the EPTS format I got an error because when compiling the regex group names can't start with a number, and for the example in the test the `player_id` is a number.
3. Taking about tests, I'll work on making them pass once we agree on the final implementation. 
4. In the metadata model I changes form having `home_team` and `away_team` to having `teams: List[Team]` with home team first and away team second, as I thin this makes it easier to work with the teams data afterwards and also allows for a third team (for example referees data sometimes present). 
5. The `Position` of players still needs working on. 
6. Before all players where listed together, now they are linked to a team, which I think is fine but just want to point this out.

Let me know what you think!

### DIFF
--- a/kloppy/domain/models/common.py
+++ b/kloppy/domain/models/common.py
@@ -45,7 +45,7 @@ class Team:
     team_id: str
     name: str
     ground: Ground
-    players: List[Player] = field(default_factory=dict)
+    players: List[Player] = field(default_factory=list)
 
     def __str__(self):
         return self.team_id
@@ -57,6 +57,13 @@ class Team:
         if not isinstance(other, Team):
             return False
         return self.team_id == other.team_id
+
+    def get_player_by_jersey_number(self, jersey_no: str):
+        for player in self.players:
+            if player.jersey_no == jersey_no:
+                return player
+
+        return None
 
 
 class BallState(Enum):
--- a/kloppy/domain/models/common.py
+++ b/kloppy/domain/models/common.py
@@ -32,11 +32,11 @@ class Position:
 class Player:
     player_id: str
     team: "Team"
-    name: str
-    first_name: str
-    last_name: str
     jersey_no: str
-    position: Position
+    name: str = None
+    first_name: str = None
+    last_name: str = None
+    position: Position = None
     attributes: Optional[Dict] = field(default_factory=dict, compare=False)
 
 
--- a/kloppy/domain/models/common.py
+++ b/kloppy/domain/models/common.py
@@ -18,14 +18,14 @@ class Ground(Enum):
     REFEREE = "referee"
 
 
-@dataclass
+@dataclass(frozen=True)
 class Position:
     position_id: str
     name: str
     coordinates: Point
 
 
-@dataclass
+@dataclass(frozen=True)
 class Player:
     player_id: str
     team: "Team"
@@ -34,7 +34,7 @@ class Player:
     last_name: str
     jersey_no: str
     position: Position
-    attributes: Optional[Dict] = field(default_factory=dict)
+    attributes: Optional[Dict] = field(default_factory=dict, compare=False)
 
 
 @dataclass
@@ -46,6 +46,14 @@ class Team:
 
     def __str__(self):
         return self.team_id
+
+    def __hash__(self):
+        return hash(self.team_id)
+
+    def __eq__(self, other):
+        if not isinstance(other, Team):
+            return False
+        return self.team_id == other.team_id
 
 
 class BallState(Enum):
--- a/kloppy/domain/models/common.py
+++ b/kloppy/domain/models/common.py
@@ -17,6 +17,9 @@ class Ground(Enum):
     AWAY = "away"
     REFEREE = "referee"
 
+    def __str__(self):
+        return self.value
+
 
 @dataclass(frozen=True)
 class Position:
--- a/kloppy/domain/models/common.py
+++ b/kloppy/domain/models/common.py
@@ -146,8 +146,7 @@ class DataRecord(ABC):
 
 @dataclass
 class MetaData:
-    home_team: Team
-    away_team: Team
+    teams: List[Team]
     periods: List[Period]
     pitch_dimensions: PitchDimensions
     score: List  # first home, second away [0,0]
--- a/kloppy/domain/models/common.py
+++ b/kloppy/domain/models/common.py
@@ -90,7 +90,6 @@ class Orientation(Enum):
         attacking_direction: AttackingDirection,
         ball_owning_team: Team,
         action_executing_team: Team,
-        meta_data: "MetaData",
     ):
         if self == Orientation.FIXED_HOME_AWAY:
             return -1
@@ -111,18 +110,18 @@ class Orientation(Enum):
             else:
                 raise Exception("AttackingDirection not set")
         elif self == Orientation.BALL_OWNING_TEAM:
-            if ball_owning_team.team_id == MetaData.teams[0].team_id:
+            if ball_owning_team.ground == Ground.HOME:
                 return -1
-            elif ball_owning_team.team_id == MetaData.teams[1].team_id:
+            elif ball_owning_team.ground == Ground.AWAY:
                 return 1
             else:
                 raise Exception(
                     f"Invalid ball_owning_team: {ball_owning_team}"
                 )
         elif self == Orientation.ACTION_EXECUTING_TEAM:
-            if action_executing_team.team_id == MetaData.teams[0].team_id:
+            if action_executing_team.ground == Ground.HOME:
                 return -1
-            elif action_executing_team.team_id == MetaData.teams[1].team_id:
+            elif action_executing_team.ground == Ground.AWAY:
                 return 1
             else:
                 raise Exception(
--- a/kloppy/domain/models/common.py
+++ b/kloppy/domain/models/common.py
@@ -6,21 +6,32 @@ from typing import Optional, List, Dict
 from .pitch import PitchDimensions
 
 
-class Team(Enum):
-    HOME = "home"
-    AWAY = "away"
-
-    def __str__(self):
-        return self.value
+@dataclass
+class Position:
+    position_id: str
+    name: str
+    coordinates: List
 
 
 @dataclass
 class Player:
     player_id: str
-    team: Team
     name: str
+    first_name: str
+    last_name: str
     jersey_no: str
-    attributes: Optional[Dict] = field(default_factory=dict)
+    position: Position
+    # attributes: Optional[Dict] = field(default_factory=dict)
+
+
+@dataclass
+class Team:
+    team_id: str
+    name: str
+    players: List[Player]
+
+    def __str__(self):
+        return self.team_id
 
 
 class BallState(Enum):
@@ -54,6 +65,7 @@ class Orientation(Enum):
         attacking_direction: AttackingDirection,
         ball_owning_team: Team,
         action_executing_team: Team,
+        meta_data: "MetaData",  # TODO is the using before assignment a problem?
     ):
         if self == Orientation.FIXED_HOME_AWAY:
             return -1
@@ -74,18 +86,18 @@ class Orientation(Enum):
             else:
                 raise Exception("AttackingDirection not set")
         elif self == Orientation.BALL_OWNING_TEAM:
-            if ball_owning_team == Team.HOME:
+            if ball_owning_team.team_id == MetaData.home_team.team_id:
                 return -1
-            elif ball_owning_team == Team.AWAY:
+            elif ball_owning_team.team_id == MetaData.away_team.team_id:
                 return 1
             else:
                 raise Exception(
                     f"Invalid ball_owning_team: {ball_owning_team}"
                 )
         elif self == Orientation.ACTION_EXECUTING_TEAM:
-            if action_executing_team == Team.HOME:
+            if action_executing_team.team_id == MetaData.home_team.team_id:
                 return -1
-            elif action_executing_team == Team.AWAY:
+            elif action_executing_team.team_id == MetaData.away_team.team_id:
                 return 1
             else:
                 raise Exception(
@@ -133,18 +145,18 @@ class DataRecord(ABC):
 
 
 @dataclass
-class Dataset(ABC):
-    flags: DatasetFlag
-    pitch_dimensions: PitchDimensions
-    orientation: Orientation
+class MetaData:
+    home_team: Team
+    away_team: Team
     periods: List[Period]
-    records: List[DataRecord]
+    pitch_dimensions: PitchDimensions
+    score: List  # first home, second away [0,0]
+    frame_rate: float
+    orientation: Orientation
+    flags: DatasetFlag
 
 
 @dataclass
-class MetaData:
-    home_team_name: str
-    away_team_name: str
-    players: List[Player]
-    periods: List[Period]
-    pitch_dimensions: PitchDimensions
+class Dataset(ABC):
+    records: List[DataRecord]
+    meta_data: MetaData
--- a/kloppy/domain/models/common.py
+++ b/kloppy/domain/models/common.py
@@ -65,6 +65,13 @@ class Team:
 
         return None
 
+    def get_player_by_id(self, id: str):
+        for player in self.players:
+            if player.player_id == id:
+                return player
+
+        return None
+
 
 class BallState(Enum):
     ALIVE = "alive"
--- a/kloppy/domain/models/common.py
+++ b/kloppy/domain/models/common.py
@@ -3,19 +3,32 @@ from dataclasses import dataclass, field
 from enum import Enum, Flag
 from typing import Optional, List, Dict
 
-from .pitch import PitchDimensions
+from .pitch import PitchDimensions, Point
+
+
+@dataclass
+class Score:
+    home: int
+    away: int
+
+
+class Ground(Enum):
+    HOME = "home"
+    AWAY = "away"
+    REFEREE = "referee"
 
 
 @dataclass
 class Position:
     position_id: str
     name: str
-    coordinates: List
+    coordinates: Point
 
 
 @dataclass
 class Player:
     player_id: str
+    team: "Team"
     name: str
     first_name: str
     last_name: str
@@ -28,7 +41,8 @@ class Player:
 class Team:
     team_id: str
     name: str
-    players: List[Player]
+    ground: Ground
+    players: List[Player] = field(default_factory=dict)
 
     def __str__(self):
         return self.team_id
@@ -65,7 +79,7 @@ class Orientation(Enum):
         attacking_direction: AttackingDirection,
         ball_owning_team: Team,
         action_executing_team: Team,
-        meta_data: "MetaData",  # TODO is the using before assignment a problem?
+        meta_data: "MetaData",
     ):
         if self == Orientation.FIXED_HOME_AWAY:
             return -1
@@ -86,18 +100,18 @@ class Orientation(Enum):
             else:
                 raise Exception("AttackingDirection not set")
         elif self == Orientation.BALL_OWNING_TEAM:
-            if ball_owning_team.team_id == MetaData.home_team.team_id:
+            if ball_owning_team.team_id == MetaData.teams[0].team_id:
                 return -1
-            elif ball_owning_team.team_id == MetaData.away_team.team_id:
+            elif ball_owning_team.team_id == MetaData.teams[1].team_id:
                 return 1
             else:
                 raise Exception(
                     f"Invalid ball_owning_team: {ball_owning_team}"
                 )
         elif self == Orientation.ACTION_EXECUTING_TEAM:
-            if action_executing_team.team_id == MetaData.home_team.team_id:
+            if action_executing_team.team_id == MetaData.teams[0].team_id:
                 return -1
-            elif action_executing_team.team_id == MetaData.away_team.team_id:
+            elif action_executing_team.team_id == MetaData.teams[1].team_id:
                 return 1
             else:
                 raise Exception(
@@ -149,7 +163,7 @@ class MetaData:
     teams: List[Team]
     periods: List[Period]
     pitch_dimensions: PitchDimensions
-    score: List  # first home, second away [0,0]
+    score: Score
     frame_rate: float
     orientation: Orientation
     flags: DatasetFlag
--- a/kloppy/domain/models/common.py
+++ b/kloppy/domain/models/common.py
@@ -21,7 +21,7 @@ class Player:
     last_name: str
     jersey_no: str
     position: Position
-    # attributes: Optional[Dict] = field(default_factory=dict)
+    attributes: Optional[Dict] = field(default_factory=dict)
 
 
 @dataclass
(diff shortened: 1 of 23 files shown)


kloppy/domain/models/common.py
```diff
@@ -54,6 +65,7 @@ def get_orientation_factor(
         attacking_direction: AttackingDirection,
         ball_owning_team: Team,
         action_executing_team: Team,
+        meta_data: "MetaData",  # TODO is the using before assignment a problem?
```
bdagnino, line 68: When referring to MetaData when type hinting I got a warning by the linter saying it it was referenced before declared. Set it as a string to type hint but avoid the warning, but not sure what is the proper way to go about this.

koenvo:
I noticed I did some detail checking of the code. I think the code is good.

1) There should be an easy way to find the home/away team. I would like to be able to have a match defined in an external system which only has a reference to some files without knowing the details (teamId, playerIds) of the data. When we have a way to go from home/away to teamId it would be fine.

2) The formats used in the EPTS serializer are not connected to the column names in the to_pandas function. We can keep the regex capture name with `player_` and remove it from to_pandas

3) of course

4) see 1

5) see comments on the code itself. Option is to rename current `Position` to `Location` to make a distinction

6) see comments on the code itself. Makes sense. A deprecated property for backwards compatibility would be good here.

kloppy/domain/models/common.py
bdagnino: Ok. Just to confirm, before my changes it was already MetaData in the models and in the models, reader and serializers of the EPTS format. Also, in the Tracab test files it's TracabMetaData. I can change it here and in the other files as well, but just wanted to confirm that's indeed what you want to do.

kloppy/domain/models/common.py
```diff
@@ -133,18 +145,17 @@ class DataRecord(ABC):
 
 
 @dataclass
-class Dataset(ABC):
-    flags: DatasetFlag
+class MetaData:
+    teams: List[Team]
```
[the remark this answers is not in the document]
bdagnino: Not sure about games with one team, but yes about games with three as you can have the referees as a third team. We have that in the past at Metrica and for example in the MLS now second spectrum also tracks the referees, so quite likely that we can encounter this again in the future. We could have them in order [home, away, referee] but we could also add a Ground field to Team.

kloppy/domain/models/common.py
```diff
@@ -6,23 +6,34 @@
...
-    HOME = "home"
-    AWAY = "away"
-
-    def __str__(self):
-        return self.value
+@dataclass
+class Position:
+    position_id: str
+    name: str
+    coordinates: List
```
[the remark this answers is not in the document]
bdagnino: Indeed it's [x,y]. I'll change it to `coordinates: Point` as it's a point on the pitch.

kloppy/domain/models/common.py
```diff
@@ -6,23 +6,34 @@
...
 
 @dataclass
 class Player:
     player_id: str
-    team: Team
     name: str
+    first_name: str
+    last_name: str
     jersey_no: str
+    position: Position
```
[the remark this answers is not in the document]
bdagnino: not sure what the issue is here. What are you referring to by *using the same name for different things*? By using position for the Player attribute and Position for the model? If so I think it makes sense and it's actually clearer that the position attribute is a Position class. Otherwise might lead to confusion.

Other than here, I don't think position is used anywhere else, or is it?

kloppy/domain/models/tracking.py
```diff
@@ -8,14 +8,12 @@
 @dataclass
 class Frame(DataRecord):
     frame_id: int
-    home_team_player_positions: Dict[str, Point]
-    away_team_player_positions: Dict[str, Point]
+    players_positions: Dict[str, Point]
```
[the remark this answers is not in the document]
bdagnino: See my previous comment about location vs position. Not sure what the conflict is.

kloppy/infra/serializers/tracking/epts/meta_data.py
```diff
@@ -123,12 +126,27 @@ def load_meta_data(meta_data_file: Readable) -> EPTSMetaData:
...
+        "away": score_elm.attrib["idVisitingTeam"],
+    }
+
+    _team_name_map = {
+        team_elm.attrib["id"]: str(team_elm.find("Name"))
+        for team_elm in meta_data.find("Teams").iterchildren(tag="Team")
+    }
+
+    teams_meta_data = {
+        ground: Team(
```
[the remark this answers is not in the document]
bdagnino: Can you elaborate a bit more on this? Not sure exactly what you mean. =/

bdagnino:
Hi Koen, 

I left some comments in the code on the things I'm not sure how to implement/change. Once I get your feedback on them I'll get to it! 

Thanks!

kloppy/domain/models/common.py
koenvo: My mistake indeed. I would like to have it Metadata everywhere. @ivd-git did a replace of data_set with dataset and would to have it consistent.

kloppy/domain/models/tracking.py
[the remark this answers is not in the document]
koenvo: I was referring to this one in the other comment. `Player.position` is of type `Position` while the values of `player_positions` are of type `Point`. I like the coordinates within `Position`; would it make sense to call it `players_coordinates` here? 
So a variable named `position` is always of type `Position` and variable named `coordinates` is always of type `Point`.

kloppy/domain/models/common.py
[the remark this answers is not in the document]
koenvo: see below

kloppy/domain/models/common.py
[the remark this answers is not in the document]
koenvo: The Ground would be good! (Altho I wasn't familiar with this term)

kloppy/infra/serializers/tracking/epts/meta_data.py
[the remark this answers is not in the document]
koenvo: as discussed above: Ground(Enum) with options HOME, AWAY and REFEREE. Instead of strings "home", "away", "referee"

kloppy/domain/models/tracking.py
```diff
@@ -1,26 +1,45 @@
...
+        }
+
+    @property
+    def away_team_players_coordinates(self):
+        return {
+            player_id: coordinates
+            for (player_id, coordinates) in self.players_coordinates.items()
+            if self.players_ground[player_id] == Ground.AWAY
+        }
+
```
bdagnino, line 30: To be able to set [home, away]_team_players_coordinates as a property, I had to add a players_ground property to the Frame class, because from the Frame class I didn't have access to meta_data.teams.players. 

It works, but I'm not sure if there is a better way to do this.

kloppy/infra/serializers/tracking/epts/meta_data.py
```diff
@@ -57,18 +67,22 @@ def _load_periods(global_config_elm, frame_rate: int) -> List[Period]:
     return periods
 
 
-def _load_players(players_elm, team_map: Dict[str, Team]) -> List[Player]:
+def _load_players(players_elm, team: Team) -> List[Player]:
     return [
         Player(
-            team=team_map[player_elm.attrib["teamId"]],
+            team=team,
```
bdagnino, line 73: Now we are adding to each Player, the full Team object (which already contains all the players). Not sure if this is correct, or if we should add only the team_id instead of the full team object.

kloppy/infra/serializers/tracking/epts/serializer.py
```diff
@@ -154,14 +149,9 @@ def deserialize(
             else None
```
bdagnino, line 149: Here Orientation is the only property of meta_data that is being set here instead that on meta_data.py, which might not be ideal and might be preferred to set everything in meta_data.py. The reasons why I dindn't do that are:

1. It will still need to check in serializer.py if it's none, and if so try to get it from the first frame (which bt might not be a good idea as sometimes the first frame is garbage)
2. Since the `orientation = (...` would be called in two places it would be better to abstract it but didn't know where would be the right place for that method to be placed.

kloppy/domain/models/common.py
bdagnino: ok! I looked at it and there are quite some places and files where it need to be changed, so I would prefer now to make it part of this pull request for now. I could do it later on on this same pull request or as another issue after this one is merged.

bdagnino:
@koenvo I think we should be close to a full implementation of the changes for the EPTS! Let me know your feedback :). Once we agree on the structure and implementation for this format we can discuss how to approach the other ones, as all of them would need to be changed to be able to use this new structure I believe.

kloppy/infra/serializers/tracking/epts/meta_data.py
```diff
@@ -57,21 +69,39 @@ def _load_periods(global_config_elm, frame_rate: int) -> List[Period]:
...
+                player_elm.find("ProviderPlayerParameters")
             ),
         )
         for player_elm in players_elm.iterchildren(tag="Player")
+        if player_elm.attrib["teamId"] == team.team_id
     ]
 
 
+def _load_position_data(parent_elm) -> Position:
+    player_provider_parameters = _load_provider_parameters(parent_elm)
```
bdagnino, line 94: The same function _load_provider_parameters get's called twice, on inside _load_position_data and then to set the Player attributes. Not sure that's the proper way to do it, but couldn't think of a better way to do it while still kipping the list comprehension structure.

kloppy/infra/serializers/tracking/epts/meta_data.py
koenvo: This will be a reference to the Team object instead of a full copy so that's fine.

kloppy/infra/serializers/tracking/epts/meta_data.py
```diff
@@ -155,23 +191,19 @@ def load_meta_data(meta_data_file: Readable) -> EPTSMetaData:
...
     frame_rate = int(meta_data.find("GlobalConfig").find("FrameRate"))
-    periods = _load_periods(meta_data.find("GlobalConfig"), frame_rate)
     pitch_dimensions = _load_pitch_dimensions(meta_data, sensors)
+    periods = _load_periods(meta_data.find("GlobalConfig"), frame_rate)
 
     return EPTSMetaData(
-        home_team_name=team_name_map[Team.HOME],
-        away_team_name=team_name_map[Team.AWAY],
-        players=players,
+        teams=[team for team in teams_meta_data.values()],
```
koenvo, line 215: teams=teams_meta_data.values()

kloppy/infra/serializers/tracking/epts/serializer.py
koenvo: When there is a way to retrieve it from the EPTS metadata we should use that, and otherwise just leave it None. Then we can add a OrientationEstimator or something that gets a couple of frames as input and returns the Orientation. That will make the estimating of the orientation more generic and can be removed from all other serializers too.

kloppy/domain/models/tracking.py
koenvo: I checked the dataclass and I think we should use the Player object itself as a key. The Player dataclass needs to be frozen.

Then you can do:
```
@property
def away_team_players_coordinates(self):
   return {player.id: coordinates for player, coordinates in self.player_coordinates.items() if player.team.ground == Ground.AWAY}
```

bdagnino: @koenvo ok! I tried to implement it but run with a potential conflict. In order to be able to use Player as a key, Team also needs to be frozen. However that might not be ideal as you might want to initialize a team without players and fill in the players later on? What do you think?

koenvo: Interesting problem. This circular reference makes a bit harder. The `frozen=True` is not required for the hash to work, you can also use `unsafe_hash=True`. In the meanwhile I'll check if there is a clean way to do it using `frozen`.

bdagnino: Tried using `unsafe_hash=True` for Team while it works for the previous error now I get another error, this time `TypeError: unhashable type: 'list'`. From what I was able to understand, the problem here is that `players` property of Teams is a list of players and list can't be used as key (tuples can, but not lists). The solution would be to make `Team.players` a tuple, but I think that also can bring more problems than solutions, as there are probably cases in which you need to add or remove players, to a Team after initializing it, don't you think?

Another option may be could be to pass a reference of the metadata to each frame object? From what I understand it would be a reference and not a copy (as when doing `player.team = Team`) so that shouldn't add any memory / performance issues?

koenvo: Didn't look at the lists.. sorry about that. Could have saved you some time there..
I think it's easier to fix this is create a custom hash function on team:
```
# see https://docs.python.org/3/reference/datamodel.html#object.__hash__

def __hash__(self):
   return hash(self.team_id)

def __eq__(self, other):
   if not isinstance(other, Team):
      return False
    return self.team_id == other.team_id
```

This way all other properties are ignored, which makes sense in this case.

Shouldn't we just get rid of the `away_team_players_coordinates` and `home_team_players_coordinates` properties? And make sure we have `Player` objects as keys. That would make it very easy for end-users to filter on home/away team - and other properties.

bdagnino: Alright! Have a look at the updated code, comments about what needed to be done to make Player as a key worked are there. One thing that may be is relevant, is that  using player as a key increased the time to `load_epts_tracking_data` by ~50%. From 20s to 30s con my computer.

koenvo: Cool that’s working!

When performance is really an issue we can optimize it. Have already some ideas. So no worries for now

kloppy/infra/serializers/tracking/epts/meta_data.py
bdagnino: @koenvo just to be sure that you saw this, any feedback here or it's ok as it?

koenvo: Good point! Besides calling the function twice, the ‘Player.attributes’ property should not contain the position data. But I don’t know a clean way to split it right now. If you can just add a todo / note it’s fine, or if you figure at a nice way it’s even better :-)

bdagnino:
@koenvo here go the latest changes! I wrote details about the changes on the commit messages,but let me know if you have any questions. Main points are three:

1. All serializers and helpers methods involved work with the new Dataset and MetaData models. 
2. For all but EPTS, at the moment there are no players created for each Team, to Team.players == []. I could work on it in the coming days, or we could leave it as an issue for a future improvement.
3. All test pass at the moment, but not sure if this changes brake code not covered by the tests, as I'm not super familiar with it (like examples, pattern matching for the events, etc)

Let me know what you think!

kloppy/domain/models/common.py
```diff
@@ -54,6 +90,7 @@ def get_orientation_factor(
         attacking_direction: AttackingDirection,
         ball_owning_team: Team,
         action_executing_team: Team,
+        meta_data: "MetaData",
```
koenvo, line 93: You can get rid of this one

kloppy/domain/models/common.py
```diff
@@ -74,18 +111,18 @@ def get_orientation_factor(
             else:
                 raise Exception("AttackingDirection not set")
         elif self == Orientation.BALL_OWNING_TEAM:
-            if ball_owning_team == Team.HOME:
+            if ball_owning_team.team_id == MetaData.teams[0].team_id:
```
koenvo, line 114: ball_owning_team.ground == Ground.HOME

kloppy/domain/models/common.py
```diff
@@ -74,18 +111,18 @@ def get_orientation_factor(
...
                 return -1
-            elif ball_owning_team == Team.AWAY:
+            elif ball_owning_team.team_id == MetaData.teams[1].team_id:
```
koenvo, line 116: `ball_owning_team.ground == Ground.AWAY`

kloppy/domain/models/common.py
```diff
@@ -74,18 +111,18 @@ def get_orientation_factor(
...
                 return 1
             else:
                 raise Exception(
                     f"Invalid ball_owning_team: {ball_owning_team}"
                 )
         elif self == Orientation.ACTION_EXECUTING_TEAM:
-            if action_executing_team == Team.HOME:
+            if action_executing_team.team_id == MetaData.teams[0].team_id:
```
koenvo, line 123: `action_executing_team.ground == Ground.HOME`

kloppy/domain/models/common.py
```diff
@@ -74,18 +111,18 @@ def get_orientation_factor(
...
                 return -1
-            elif action_executing_team == Team.AWAY:
+            elif action_executing_team.team_id == MetaData.teams[1].team_id:
```
koenvo, line 125: `action_executing_team.ground == Ground.AWAY`

kloppy/domain/services/__init__.py
```diff
@@ -17,16 +17,16 @@ def attacking_direction_from_frame(frame: Frame) -> AttackingDirection:
     """ This method should only be called for the first frame of a """
     avg_x_home = avg(
         [
-            player.x
-            for player in frame.home_team_player_positions.values()
-            if player
+            coordinates.x
+            for key, coordinates in frame.players_coordinates.items()
+            if "home" in key
```
koenvo, line 22: I think this should be:

```python
for player, coordinates in frame.players_coordinates.items()
            if player.team.ground == Ground.HOME
```

bdagnino: See comment on the latest commit. 

As mentioned in the previous comment of the pull request, except for the EPTS format, the list of players for each team is empty. So till that is done, we need to check if "home" or "away" are on the key of the players_coordinates.

kloppy/infra/serializers/tracking/metrica.py
```diff
@@ -175,6 +179,11 @@ def deserialize(
         # consider reading this from data
         frame_rate = 25
 
+        # TODO: also used in Tracab, extract to a method
+        home_team = Team(team_id="home", name="home", ground=Ground.HOME)
+        away_team = Team(team_id="away", name="away", ground=Ground.AWAY)
+        teams = [home_team, away_team]
```
bdagnino, line 185: @koenvo I looked a bit into creating the Players and adding them to the Team for Metrica files, and probably the same applies to Tracab. But I have a doubt regarding how to do it. 

To create the players I need to access the first two rows of the file, but as the file is consumed in the iterator, I don't know how to get them out to create the players for each team without adding them to all the frames. 

I could read the two first lines outside of `__create_iterator` but that's probably not the right way to do it? I think I should have time this week to add all the players to all other data formats other than EPTS, but might need a bit of help finding out the best way to do it.

kloppy/domain/services/__init__.py
bdagnino: Implemented.

kloppy/infra/serializers/tracking/metrica.py
bdagnino: Implemented.

bdagnino:
@koenvo I had some time to work on this today so went ahead and did it. The players information is now available on the dataset for Metrica and Tracab. I'm not sure about the implementation for Tracab (see comments on the commit).

I did not implemented it for Opta and Statsbomb serializer because I wasn't sure what you wanted to do for events. At the moments the events have a field for `jersey_no` which combined with the team you can use to get the events of a player. I suppose we should add all Player to each Team, but then not sure what to do for each event. Some options are:

1. Leave jersey_no and add Player. 
2. Remove jersey_no and add Player
3. Either one of the above using only player_id instead of Player. 

If you have an opinion on what would be best at the moment I can implement that and that would be the last things to add to this pull request (except for revisions to the code of course.) 

Let me know!

bdagnino:
> @koenvo I had some time to work on this today so went ahead and did it. The players information is now available on the dataset for Metrica and Tracab. I'm not sure about the implementation for Tracab (see comments on the commit).
> 
> I did not implemented it for Opta and Statsbomb serializer because I wasn't sure what you wanted to do for events. At the moments the events have a field for `jersey_no` which combined with the team you can use to get the events of a player. I suppose we should add all Player to each Team, but then not sure what to do for each event. Some options are:
> 
> 1. Leave jersey_no and add Player.
> 2. Remove jersey_no and add Player
> 3. Either one of the above using only player_id instead of Player.
> 
> If you have an opinion on what would be best at the moment I can implement that and that would be the last things to add to this pull request (except for revisions to the code of course.)
> 
> Let me know!

After looking at the code a bit more, I'd propose the following:

- for the record/event, leave `team = Team`, and change `player_jersey_no` for `player = Player`
- then in to_pandas, change `team=event.team.value` for `team=event.team.team_id` and `player_jersey_no=event.player_jersey_no` for `player_id=event.player.player_id`

What do you think?

bdagnino:
@koenvo I went ahead and implemented what I proposed in the last comment. I also changes the references to locations by coordinates, this wasn't done before. The details of the changes are:

1. Players created and added to teams in meta_data
2. In the Event object, players are referred by their object, meaning `player=Player` and then in the to_pandas helper, they are referred by their player_id, and the teams by their team_id
3. To make it consistent with the tracking naming, I changed all references to `location` in events to `coordinates`

I think pending suggestions or refactoring to the code, these changes complete this pull request, so we should be quite close to merging it!

koenvo:
> @koenvo I went ahead and implemented what I proposed in the last comment. I also changes the references to locations by coordinates, this wasn't done before. The details of the changes are:
> 
> 1. Players created and added to teams in meta_data
> 2. In the Event object, players are referred by their object, meaning `player=Player` and then in the to_pandas helper, they are referred by their player_id, and the teams by their team_id
> 3. To make it consistent with the tracking naming, I changed all references to `location` in events to `coordinates`
> 
> I think pending suggestions or refactoring to the code, these changes complete this pull request, so we should be quite close to merging it!

This sounds great! I'll have a look at it and see if it can be merged

kloppy/helpers.py
```diff
@@ -129,23 +130,22 @@ def _frame_to_pandas_row_converter(frame: Frame) -> Dict:
...
-    for jersey_no, position in frame.away_team_player_positions.items():
-        row.update(
-            {
-                f"player_away_{jersey_no}_x": position.x,
-                f"player_away_{jersey_no}_y": position.y,
-            }
-        )
+    for player, coordinates in frame.players_coordinates.items():
+
+        if isinstance(player, Player):
```
koenvo, line 138: For what case is this check needed? I would assume it's always an Player instance

bdagnino: This was before all data providers had Player as key in players_coordinates. Not it can be removed.

PR #34, opened 2020-07-23
Add metadata

koenvo:


kloppy/helpers.py
```diff
@@ -164,29 +158,29 @@ def _event_to_pandas_row_converter(event: Event) -> Dict:
...
-        ball_owning_team=event.ball_owning_team.value
+        ball_owning_team=event.ball_owning_team.team_id
         if event.ball_owning_team
         else None,
-        team=event.team.value,
-        player_jersey_no=event.player_jersey_no,
-        position_x=event.position.x if event.position else None,
-        position_y=event.position.y if event.position else None,
+        team=event.team.team_id,
+        player=event.player.player_id,
```
koenvo, line 165: I'm thinking this should be `player_id=event.player.player_id`, otherwise people might think you can pass a `Player` instance when filtering the dataframe

bdagnino: @koenvo agreed. Should we do that also for team and change `team=event.team.team_id` to `team_id=event.team.team_id` ?

PR #37, opened 2020-07-28
Add windows

koenvo:


kloppy/domain/models/window.py
```diff
@@ -0,0 +1,71 @@
...
+            score = Score(home=self.score.home, away=self.score.away + 1)
+        else:
+            raise Exception(f"Unknown ground {team.ground}")
+
+        return State(
+            players=self.players,
+            score=score
+        )
+
+    def add_away_goal(self) -> 'State':
```
koenvo, line 38: Remove this

kloppy/domain/services/windowers/__init__.py
```diff
@@ -0,0 +1,88 @@
...
+            key=key(current_state)
+        )
+        windows.append(current_window)
+
+        previous_event: Union[Event, None] = None
+        previous_key = object()
+
+        for event in dataset.events:
+            if previous_event and previous_event.period != event.period:
+                # On period change always state a new window
```
koenvo, line 57: Typo: 'start' instead of 'state'

bdagnino:
@koenvo I like the implementation! Some feedback:

- One thing to keep in mind  (if you think this is relevant) is how the implementation would work for other providers like Metrica, or Opta. For example at Metrica we don't have substitutions events, or starting eleven as part of the events we provide. The formation information at the moment needs to ne inferred from the EPTS metadata file. For Metrica in particular we (metrica) could think about adding this information to the new events json file. However this is not a comment in particular for Metirca, but more in general, to think how this would work for other providers. 

- One interesting concept would be fo have a Formations dataclass, that has the formation for each team, so that you can use it to define the windows, but also to query the formation at a certain timestamp/frame.

PR #38, opened 2020-07-29
added provider to metadata

FCrSTATS:
added 'provider' to the metadata of all datasets.  statsbomb, tracab, metrica, opta as current options


kloppy/helpers.py
```diff
@@ -14,6 +14,7 @@
     TrackingDataset,
     Transformer,
     Orientation,
+    Provider,
```
koenvo, line 17: Import not needed here

kloppy/infra/serializers/tracking/epts/metadata.py
```diff
@@ -242,5 +242,6 @@ def load_metadata(metadata_file: Readable) -> EPTSMetadata:
         sensors=sensors,
         score=score,
         orientation=None,
+        provider=None,
```
koenvo, line 245: @bdagnino what should we do here? EPTS isn't a provider

kloppy/infra/serializers/tracking/epts/serializer.py
```diff
@@ -9,6 +9,7 @@
     Point,
     Team,
     Orientation,
+    Provider,
```
koenvo, line 12: Import not needed

kloppy/infra/serializers/tracking/epts/metadata.py
bdagnino: @koenvo I think it would be good to have the provider as an input to load_metadata with a default to None, like:

```python
def load_metadata(metadata_file: Readable, procider: Provider = None) -> EPTSMetadata:
```

At the moment it assumes it's a Metrica data file (because of ProviderParameters it uses), but eventually it will need to handle different ProviderParameters, depending on the provider, so it'd be good to already accept this. 

What do you think?

FCrSTATS: I agree with that, I have added the default = None to epts/metadata.py 

I think it is all ready for final review and hopefully merge

koenvo: Until we have a better way to determine the provider of a EPTS lets go with the None value.  I'll merge the request. Thanks @FCrSTATS

PR #44, opened 2020-08-01
Added serializer for Metrica event data in json format

bdagnino:
In this commit we added the serializer and corresponding tests for reading Metrica Sports event data in json format. The types and subtypes provided by Metrica were used to extract the common events user for Opta and Statsbomb event data [PASS, SHOT, CARRY, TAKE ON]. All the rest are Generic Events.

Regarding our previous discussion of adding `SET PIECE` as a 5th common event. I looked into it, but in Statsbomb data Set Pieces are inside passes. Thus, adding them as an event would require changing the definition of PASS for the Statsbomb serializer, and the same for opta. I thought it would be best to keep it as it is for now.


kloppy/infra/serializers/event/metrica/json_serializer.py
```diff
@@ -0,0 +1,311 @@
...
+                    )
+
+                player = team.get_player_by_id(raw_event["from"]["id"])
+                event_type = raw_event["type"]["id"]
+                subtypes = _parse_subtypes(raw_event)
+                period = [
+                    period
+                    for period in metadata.periods
+                    if period.id == raw_event["period"]
+                ][0]
```
bdagnino, line 253: @koenvo Is there a better way to do this than indexing zero at the end of the comprehension?

kloppy/infra/serializers/event/metrica/json_serializer.py
```diff
@@ -0,0 +1,311 @@
...
+                >>> with open("events.json", "rb") as event_data, \
+                >>>      open("metadata.xml", "rb") as metadata:
+                >>>
+                >>>     dataset = serializer.deserialize(
+                >>>         inputs={
+                >>>             'event_data': event_data,
+                >>>             'metadata': metadata
+                >>>         },
+                >>>         options={
+                >>>             'event_types': ["pass", "take_on", "carry", "shot"]
```
eujern, line 214: Outside the scope of this PR, and looks like it's how it's done for the opta and statsbomb serializers currently, but is there a reason why this shouldn't just be a list of EventType enums?

kloppy/infra/serializers/event/metrica/json_serializer.py
```diff
@@ -0,0 +1,311 @@
...
+                    ball_state=BallState.ALIVE,
+                    # from Event
+                    event_id=None,
+                    team=team,
+                    player=player,
+                    coordinates=(_parse_coordinates(raw_event["start"])),
+                    raw_event=raw_event,
+                )
+
+                if event_type in MS_PASS_TYPES:
```
eujern, line 269: Could consider adding a type check here so you can skip parsing completely if the event type was not listed in the options.
```
event = None
if event_type in MS_PASS_TYPES:
    if EventType.PASS not in wanted_event_types:
        continue
...
```
line 232 would need to change to:
`for event_type in options.get("event_types", [e for e in EventType])`
and line 302 changed to:
```
if event:
    events.append(event)
```

kloppy/infra/serializers/event/metrica/json_serializer.py
koenvo: Might feed it to https://github.com/PySport/kloppy/pull/44#pullrequestreview-459658584 . But when it’s not found it should raise an exception

kloppy/infra/serializers/event/metrica/json_serializer.py
koenvo: Yes. Better to pass the EventTypes here and do the resolve from string to EventType in the helper function. I don’t want to bother users with the EventTypes because that takes additional imports.

kloppy/infra/serializers/event/metrica/json_serializer.py
koenvo: Personally I like to have a less as branches (if ... ) as possible for readability. In case when performance really suffers it might be good do skip parsing. That's not the case here.

Furthermore keeping the same structure for each serializer allows us to do a refactor and extract common parts to a generic event serializer / base event serializer.

PR #49, opened 2020-08-04
Infer ball_owning_team from Opta events

rjtavares:
Certain event types were defined as `BALL_OWNING_EVENTS` (discussion is needed over which types should be considered). Other events keep the `ball_owning_team` of the previous event.

### DIFF
--- a/kloppy/tests/test_opta.py
+++ b/kloppy/tests/test_opta.py
@@ -28,6 +28,7 @@ class TestOpta:
         assert dataset.metadata.provider == Provider.OPTA
         assert len(dataset.events) == 17
         assert len(dataset.metadata.periods) == 2
+        assert dataset.events[10].ball_owning_team == dataset.metadata.teams[1]
         assert dataset.events[15].ball_owning_team == dataset.metadata.teams[0]
         assert (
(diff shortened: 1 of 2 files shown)


koenvo:
Looks good @rjtavares. Can you update the tests to make sure we also test this new behaviour?

rjtavares:
Added a test. No idea what the 5 failing checks mean, can you help @koenvo ?

kloppy/tests/test_opta.py
```diff
@@ -28,6 +28,7 @@ def test_correct_deserialization(self):
         assert dataset.metadata.provider == Provider.OPTA
         assert len(dataset.events) == 17
         assert len(dataset.metadata.periods) == 2
+        assert dataset.events[15].ball_owning_team == dataset.metadata.teams[0]
```
koenvo, line 31: Can you update this a little bit so we know:
1. It works for events that are in the list of `BALL_OWNING_EVENTS`
2. It works for events that are not in the list of `BALL_OWNING_EVENTS`

koenvo:
One more little comment and then we can merge it

rjtavares:
Changed the test so it tests both types of events in the test file (pass and generic)

Issue #45, opened 2020-08-03
Adding a data type type to the datasets


bdagnino:
@koenvo I'm working with games in which I have a tracking dataset and an event dataset, and I'd like to have share methods, like for example `get_team_dataframe`  which should do different things depending on of the dataset is a tracking dataset or an event one. 

I'd propose we add a type DataType and a data_type attribute to the Datasets:

```python
class DataType(Enum):
    TRACKING = "tracking"
    EVENT = "event"

@dataclass
class Dataset(ABC):
    records: List[DataRecord]
    metadata: Metadata
    data_type: DataType
```

If you agree with this, I can implement it.

bdagnino:
Another option could be to include it inside `metadata`. I think since this is descriptive of the dataset it belongs to the dataset though, but don't have a high conviction on that.

bdagnino:
Closing issue as I realized I don't need this actually for now at least. Happy to do it still if relevant, but don't want to leave the issue open if not needed.

koenvo:
I reopen this because it’s a good first issue and like to have more of those.

Implementation should be similar to https://github.com/PySport/kloppy/blob/master/kloppy/domain/models/event.py#L82

bdagnino:
Perfect! I'll remove myself from this issue then. But happy to help whomever ends up implementing it if needed.

eujern:
May I take a stab at this one?

bdagnino:
@eujern yes! Let me know if you need any help or guidance.

eujern:
Actually, have a question while working on this.  `Dataset` is an abstract class, and we have `EventDataset` and `TrackingDataset` implement it.  Why is it still necessary to have an explicit `data_type` field?  For the `get_team_dataframe` example above, wouldn't you just add an abstract method to `Dataset`, and then implement them separately in `EventDataset` and `TrackingDataset` for event and tracking data respectively?

eujern:
Or if the method should not be a part of `Dataset`, couldn't you just an `isinstance(dataset, EventDataset)` check in the method?

koenvo:
> Actually, have a question while working on this. `Dataset` is an abstract class, and we have `EventDataset` and `TrackingDataset` implement it. Why is it still necessary to have an explicit `data_type` field? For the `get_team_dataframe` example above, wouldn't you just add an abstract method to `Dataset`, and then implement them separately in `EventDataset` and `TrackingDataset` for event and tracking data respectively?

[...]

PR #50, opened 2020-08-07
Add DatasetType.

eujern:
#45 Adding a data type type to the datasets

### DIFF
--- a/kloppy/domain/models/common.py
+++ b/kloppy/domain/models/common.py
@@ -1,4 +1,4 @@
-from abc import ABC
+from abc import ABC, abstractmethod
 from dataclasses import dataclass, field
 from enum import Enum, Flag
 from typing import Optional, List, Dict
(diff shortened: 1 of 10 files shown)


kloppy/domain/models/common.py
```diff
@@ -224,7 +224,17 @@ class Metadata:
     provider: Provider
 
 
+class DatasetType(Enum):
```
eujern, line 227: Same naming convention as `EventType`.  There is however `DatasetType = TypeVar("DatasetType")` defined in line 118 of `kloppy/helpers.py` as well as `kloppy/domain/services/transformers/__init__.py`.  In the latter case, it's class scoped.  Could be confusing. Necessary to rename?

koenvo: Good you kept the naming convention! 

The code in `TypeVar` is used for typehinting. Would be good indeed to rename it. Let's use `DatasetT` for `helpers.py` and the transformers

eujern: Renamed.

koenvo:
Sorry I missed this one: please add an assert to the tests to make sure the value is set correctly

eujern:
Updated tests.

koenvo:
Great!

Issue #46, opened 2020-08-03
Read EPTS data provider from metadata.xml file


bdagnino:
@koenvo now you can set the provider of the EPTS data from `load_metadata`, however you can't do it from the load helper method `load_epts_tracking_data`. Wold be nice to have it there as well so you can use this helper without having to set it after the dataset is created. 

I can implement this if you agree.

koenvo:
Can we detect it from the file itself? Even an ugly way (like checking existence of certain values in the xml)

 When a user has to specify it I’m not sure I see the goal of it.

bdagnino:
Yes, you are right. Moreover, there is actually an attribute  in the file for provider `<ProviderName>Example</ProviderName>` so we can use that to check for known providers. Do you want to leave this one also as a good first issue? I'm in no urgency for this yet.

TK5-Tim:
@bdagnino So basically `load_epts_tracking_data` should work similar to `load_meta_data`?

bdagnino:
@TK5-Tim not exactly. That's what I originally proposed, but @koenvo suggested a better approach. His proposal is to read the provider from the ETPS metadata file itself when loading the metadata. For that the following background is important:

1. If you look [here](https://github.com/PySport/kloppy/blob/1c9a62948a43313130be52bc5874f573d20b5d8b/kloppy/domain/models/common.py#L24) you'll see there are 4 providers, for now the only one that metters for the EPTS sreializer is Metrica, because we are the only ones providing data in the EPTS format (at least that we have access to), but this is relevant for this issue. 

2. Then you'll see that [load_metadata](https://github.com/PySport/kloppy/blob/1c9a62948a43313130be52bc5874f573d20b5d8b/kloppy/infra/serializers/tracking/epts/metadata.py#L153) takes 2 arguments, the `metadata_file` and the provider if known. In this case the provider won't be known beforehand, because it's declared in the metadata file. 

3. The metadata file, has provider name attribute. In the example file that's on the tests folder of Kloppy it says `<ProviderName>Example</ProviderName>`, in files from Metrica is says `<ProviderName>Metrica Sports</ProviderName>`.

So the implementation would be to inside `load_metadata` read the `ProviderName` attribute and if it matches `Metrica Sports`, then set the provider to Provider.METRICA, and otherwise leave it as None. Later on when we get more EPTS files from other providers, we'll need to add there other cases as well. 

[...]

PR #52, opened 2020-08-09
#46 read epts data provider from metadata file

TK5-Tim:
Hey @bdagnino 

I got a first proposal for the implementation of #46 
Sorry for the delay I was taking a too complicated approach at first. 

What I implemented: 
* Reading the Provider name from the XML similar to how the Score is read from the file.
* Raise a warning if the Provider name given with the method is different to the provider name in the file. 
* change the provider variable to Provider.METRICA if Metrica is set as provider in the xml, leave it as it is otherwise 
* create a testing file with Metrica as ProviderName for testing 
* create a test for the provider name. 

I hope I did everything right with the test. I honestly can't figure out how to properly run them.


TK5-Tim:
So now I think I also got everything fixed.

kloppy/infra/serializers/tracking/epts/metadata.py
```diff
@@ -156,6 +157,18 @@ def load_metadata(
     root = objectify.fromstring(metadata_file.read())
     metadata = root.find("Metadata")
 
+    provider_path = objectify.ObjectPath("Metadata.GlobalConfig.ProviderName")
+    provider_name = provider_path.find(metadata)
+
+    if provider != None and provider != provider_name:
+        warnings.warn(
```
koenvo, line 164: Couple of review comments:
1. Pythonic way of checking if a variable is not None is just ‘if variable’
2. You compare a string with a Provider type, they will never be the same. A test to make sure a warning is thrown could confirm this
3. What happens when a provider name is set that we don’t know? I expect a warning
4. @bdagnino when I supply provider as an argument I expect that I overrule whatever is in the file. In current implementation I get a warning and the one in the file is leading. What do you think that should happen?

kloppy/tests/files/epts_metrica.xml
```diff
@@ -0,0 +1,306 @@
+<?xml version="1.1" encoding="UTF-8"?>
+<tns:FIFADataTransferFormatEPTS
+	xmlns:tns="FIFADataTransferFormatEPTSNamespace/FIFADataTransferFormatEPTS"
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
```
koenvo, line 4: Please delete as possible from the xml to let the test still pass. This makes it easier find problems when the test breaks.

kloppy/tests/test_epts.py
```diff
@@ -41,6 +42,13 @@ def test_regex(self):
 
         assert result is not None
 
+    def test_provider_name_recognition(self):
+        base_dir = os.path.dirname(__file__)
+        with open(f"{base_dir}/files/epts_metrica.xml", "rb") as metadata_fp:
+            metadata = load_metadata(metadata_fp)
+
+        assert metadata.provider is Provider.METRICA
```
koenvo, line 50: This should be ‘==‘ instead of ‘is’. ‘is’ checks if variables point to the same object while ‘==‘ checks if they have the same value. In this case we need check values.

kloppy/tests/files/epts_metrica.xml
TK5-Tim: I think there is something missing. Did you want to say as much as possible?

kloppy/infra/serializers/tracking/epts/metadata.py
TK5-Tim: I slipped up with 2. in the test aswell so yeah that totally makes sense.

TK5-Tim:
Thanks @koenvo for the feedback. I will try to get to it sometime later today after work 😄

kloppy/infra/serializers/tracking/epts/metadata.py
TK5-Tim: To compare them right do we need like a dictionary of accepted names in the XML? because we have a different name in the Provider class enum than it would be in the XML. And where is the right place to put that dictionary?

kloppy/tests/files/epts_metrica.xml
koenvo: Ah yes. As much as possible 👍

kloppy/infra/serializers/tracking/epts/metadata.py
koenvo: For now this could be just be a 
```python
if AAA

elif BBB

else:
  raise Exception(f"Unknown provider {provider_name}. Please create a ticket at https://github.com/PySport/kloppy/issues")
```

TK5-Tim: This should be fine! Are you also fine with creating a separate function?

koenvo: Absolutely!

bdagnino: @koenvo about point 4, I agree the one given as input should prevail. Regarding the warning, probably ok to leave it? Could be a bit annoying if you really know what you are doing, but on the other hand I can't think of a good reason to want to do that (input a different provider than the one on the file).

koenvo: Agree and let's go with the warning. People might read it :-)

kloppy/infra/serializers/tracking/epts/metadata.py
```diff
@@ -151,23 +151,42 @@ def _load_pitch_dimensions(
         return None
 
 
+def _load_provider(metadata_elm, provider: Provider) -> Provider:
+    provider_path = objectify.ObjectPath("Metadata.GlobalConfig.ProviderName")
+    provider_name = provider_path.find(metadata)
+
+    if provider:
```
koenvo, line 158: I would do the parsing just once. 
1. Read and parse provider from file into ‘provider_from_file’
2. If ‘provider’ is set, only compare with the on from file and raise warning if needed
3. Else (‘provider’ is not set), set ‘provider’ to value from file
4. Return provider

And since we only support python 3.6 please use f-strings for formatting

koenvo:
Make sure the tests run successfully :-)

TK5-Tim:
@koenvo I got another string of commits together with whom I am pretty confident it's an improvement. But I am also very much open for more Feedback :)

kloppy/infra/serializers/tracking/epts/metadata.py
```diff
@@ -150,12 +151,41 @@ def _load_pitch_dimensions(
...
+    return provider
+
+
 def load_metadata(
     metadata_file: Readable, provider: Provider = None
 ) -> EPTSMetadata:
     root = objectify.fromstring(metadata_file.read())
     metadata = root.find("Metadata")
 
+    provider = _load_provider(metadata)
```
koenvo, line 187: Please pass original `provider` argument as second argument to `_load_provider`

PR #55, opened 2020-08-27
Add StateBuilder

koenvo:


kloppy/domain/services/state_builder/lineup.py
```diff
@@ -0,0 +1,58 @@
...
+            )
+        )
+
+    def reduce(self, state: Lineup, event: Event) -> Lineup:
+        if isinstance(event, SubstitutionEvent):
+            state = Lineup(
+                players=state.players - {event.player}
+                | {event.replacement_player}
+            )
+        elif isinstance(event, PlayerOffEvent):
```
eujern, line 51: My apologies for not being familiar with the dataset formats, but is the differentiation between SubstitutionEvent and PlayerOn/Off because different providers have different ways of representing substitutions?

kloppy/infra/serializers/event/statsbomb/serializer.py
```diff
@@ -40,6 +45,13 @@
 
 SB_EVENT_TYPE_HALF_START = 18
 SB_EVENT_TYPE_HALF_END = 34
+SB_EVENT_TYPE_STARTING_XI = 35
+
+SB_EVENT_TYPE_SUBSTITUTION = 19
```
eujern, line 50: Nevermind, I see statsbomb have both substition and on/off.  Will familiarize myself with the dataset formats.

kloppy/domain/services/state_builder/lineup.py
koenvo: Substitution is the official when a player is substituted. The PlayerOff/PlayerOn are the ones when a player leaves the pitch for any other reason, for injury treatment for example.

eujern: I see, thank you!

kloppy/domain/services/state_builder/__init__.py
```diff
@@ -0,0 +1,39 @@
...
+
+from kloppy.domain import List, EventDataset
+
+from .base import StateBuilder
+from .lineup import LineupStateBuilder
+from .score import ScoreStateBuilder
+from .sequence import SequenceStateBuilder
+
+
+def create_state_builder(builder_key: str):
```
deepxg, line 11: Why this instead of just using the class as key? How do I add new state builders as a consumer of the library?

kloppy/domain/services/state_builder/base.py
```diff
@@ -0,0 +1,16 @@
...
+T = TypeVar("T")
+
+
+class StateBuilder(ABC):
+    @abstractmethod
+    def initial_state(self, dataset: EventDataset) -> T:
+        pass
+
+    @abstractmethod
+    def reduce(self, state: T, event: Event) -> T:
```
deepxg, line 15: This signature makes it hard to do lookahead and lookbehind for more interesting types of state. You could hide things in state to communicate through time if you wanted to but that seems messy. You also can't build a state builder that itself relies on state from another builder.

I'd much rather see something like Clojure's transducers here - they give you the option of maintaining internal state across elements of the collection, but can also compose together to create actual pipelines.

koenvo: I'll move this to the next version.

kloppy/domain/services/state_builder/__init__.py
koenvo: Solved with a Registery.

PR #63, opened 2020-09-25
Update pitch.py

FCrSTATS:
added distance_to_point function to the Point class

it returns a int and uses base python with no additional dependencies and is the most efficient version of the calculation I managed to create

### DIFF
--- a/kloppy/domain/models/pitch.py
+++ b/kloppy/domain/models/pitch.py
@@ -26,6 +26,9 @@ class Point:
     x: float
     y: float
 
-    def distance_to_point(self, Point) -> int:
+    def distance_to_point(self, Point) -> float:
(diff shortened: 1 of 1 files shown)


kloppy/domain/models/pitch.py
```diff
@@ -25,3 +24,7 @@ class PitchDimensions:
 class Point:
     x: float
     y: float
+
+    def distance_to_point(self, Point) -> int:
```
koenvo, line 28: your definition of the argument is a bit incorrect and the return type should be float.
`def distance_to_point(self, point: Point) -> float"

and then do something with `self.x, self.y, point.x and point.y`. Hint: no zip needed. Your construction is only needed when you don't know the number of dimensions.

PR #70, opened 2020-10-22
Added improved standard events to Metrica serializer

bdagnino:
As part of the larger issue of adding more standard event to be able to
better match the data format and event types between provier, i have
aded in this commit the following standard events:

	- GOAL KICK
	- THROW IN
	- FREE KICK
	- CORNER KICK
	- KICK OFF
	- PENALTY
	- RECOVERY
	- FOUL COMMITTED
	- BALL OUT

In the case of Metrica data, BALL OUT is a synthetic event that's added
to the dataset. All the other ones replace the orginal event, for
example, before a pass froma free kick was of event_type PASS, now it
woul dbe of event_type FREE KICK.

I just realized shots from penaly or free kicks are now correctly
handled, will fixed that in a next commit.

The same changes will also implemented in the other providers.

### DIFF
--- a/kloppy/domain/models/event.py
+++ b/kloppy/domain/models/event.py
@@ -76,6 +76,15 @@ class EventType(Enum):
     CARD = "CARD"
     PLAYER_ON = "PLAYER_ON"
     PLAYER_OFF = "PLAYER_OFF"
+    RECOVERY = "RECOVERY"
+    BALL_OUT = "BALL_OUT"
+    FOUL_COMMITTED = "FOUL_COMMITTED"
+    GOAL_KICK = "GOAL_KICK"
+    FREE_KICK = "FREE_KICK"
+    THROW_IN = "THROW_IN"
+    CORNER_KICK = "CORNER_KICK"
+    PENALTY = "PENALTY"
+    KICK_OFF = "KICK_OFF"
 
 
 @dataclass
@@ -179,6 +188,24 @@ class CardEvent(Event):
 
 
 @dataclass
+class RecoveryEvent(Event):
+    event_type: EventType = EventType.RECOVERY
+    event_name: str = "recovery"
+
+
+@dataclass
+class BallOutEvent(Event):
+    event_type: EventType = EventType.BALL_OUT
+    event_name: str = "ball_out"
+
+
+@dataclass
+class FoulCommittedEvent(Event):
+    event_type: EventType = EventType.FOUL_COMMITTED
+    event_name: str = "foul_committed"
+
+
+@dataclass
 class EventDataset(Dataset):
     records: List[
         Union[
@@ -225,4 +252,7 @@ __all__ = [
     "CardEvent",
     "CardType",
     "EventDataset",
+    "RecoveryEvent",
+    "FoulCommittedEvent",
+    "BallOutEvent",
 ]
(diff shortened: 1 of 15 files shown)


bdagnino:
@koenvo still need to do some work and see how to resolve some conflict around free kicks and penalties, but creating this PR in case you want to have a look at the implementation so far. Let me know if you have any comments.

kloppy/helpers.py
```diff
@@ -202,6 +202,12 @@ def _event_to_pandas_row_converter(event: Event) -> Dict:
                 "end_coordinates_y": event.end_coordinates.y,
             }
         )
+
+    if event.qualifiers:
+        for qualifier in event.qualifiers:
+            if isinstance(qualifier, SetPieceType):
+                row.update({f"is_{qualifier.name}": True})
```
bdagnino, line 209: @koenvo not sure about this implementation. Another option could be to define the value of the enum SetPieceType to True, so that you can always do name + value pairs no matter the instance type. I'm thinkin about having Foot for instance as qualifiers as you mentioned or other qualifiers where the value is not necessarily True or False.

bdagnino:
@koenvo can you / do you want to check this new implementation out so that we go with it or modify it before applying it to other providers?

Besides the comment left on the code, another thing I'm not sure is how to define a parent category that includes all possible Qualifiers, while still having independent enums for each one of them. For example, if we have SetPieceType and Foot. Each one is an enum, but how can we know they are both Qualifiers?

Let me know what you think of the implementation!

kloppy/helpers.py
koenvo: Looks like the right direction. I tried to make it a bit more generic:

```python

@dataclass
class Qualifier(ABC):
    @abstractmethod
    def to_dict(self):
        pass

    @property
    def name(self):
        return camelcase_to_snakecase(
            removesuffix(type(self).__name__, 'Qualifier')
        )


class BoolQualifier(Qualifier, ABC):
    value: bool

    def to_dict(self):
        return {
            f'is_{self.name}': self.value
        }


class EnumQualifier(Qualifier, ABC):
    value: Enum

    def to_dict(self):
        return {
            f'{self.name}_type': self.value.value
        }


class SetPieceType(Enum):
    GOAL_KICK = "goal_kick"
    CORNER_KICK = "corner_kick"


@dataclass
class SetPieceQualifier(EnumQualifier):
    value: SetPieceType


a = SetPieceQualifier(value=SetPieceType.CORNER_KICK)
bla = a.to_dict()
print(bla)


class Event:
    qualifiers: List[Qualifier]

```

bdagnino: Niiice! Exactly what I was thinking of but didn't know how to implement. Thanks!

bdagnino:
@koenvo we can discuss some details tomorrow on the meeting, but pending review and may be some light refactoring this should be close to merge-ready  :)

kloppy/domain/models/event.py
```diff
@@ -90,6 +136,8 @@ class Event(DataRecord, ABC):
     raw_event: Dict
     state: Dict[str, any]
 
+    qualifiers: List[SetPieceType]
```
koenvo, line 139: `List[Qualifier]`

kloppy/utils.py
```diff
@@ -44,3 +44,7 @@ def camelcase_to_snakecase(name):
     """Convert camel-case string to snake-case."""
     s1 = _first_cap_re.sub(r"\1_\2", name)
     return _all_cap_re.sub(r"\1_\2", s1).lower()
+
+
+def removes_suffix(string, suffix):
```
koenvo, line 49: Maybe check if suffix is in string?

PR #71, opened 2020-11-05
Sportec serializer

koenvo:
TODO:
- [X] Read metadata
- [x] Read event data
  - [x] Read orientation
- [x] Strip Sportec files to use in automated tests


kloppy/infra/serializers/event/sportec/serializer.py
```diff
@@ -0,0 +1,479 @@
...
+                if (
+                    SPORTEC_EVENT_NAME_KICKOFF in event_chain
+                    and "GameSection"
+                    in event_chain[SPORTEC_EVENT_NAME_KICKOFF]
+                ):
+                    period_id += 1
+                    period = Period(
+                        id=period_id,
+                        start_timestamp=timestamp,
+                        end_timestamp=None,
```
bdagnino, line 301: I understand there is no end time for the periods? May be using the start time of the next one as end of the previous one? Or better leave that to the user to patch if needed?

bdagnino:
@koenvo looks good to me! Only concern is that may be reading the previous event in the list to get information like ball out or pass end coordinantes could not be accurate? It assumes the events are always in the correct chronological order. 

I think this is well handled in the code and it's better having it like this that not having it, but just wanted to mention it in case you have some extra info or thoughts on the topic.

kloppy/infra/serializers/event/sportec/serializer.py
koenvo: The period is ended by the 'FinalWhistle' event. So until an half is ended the period.end_timestamp will be None. But it will be updated .

koenvo:
Hmm yeah I assume data is sorted on timestamp. I've seen some events that had the same timestamp. But it will always be increasing or the same. When it becomes an issue I can sort it.

Issue #72, opened 2020-11-09
Improve sequence definition


bdagnino:
As discussed, I propose we improve the definition of sequences. Currently to start a new sequence, only the team of an event is checked: 

```python
if state.team != event.team:
    state = replace(
        state, sequence_id=state.sequence_id + 1, team=event.team
    )
```

The problem with this definition is that you can have events that don't define a sequence define one. Plus, a sequence only ends with an event from another team, while for example the ball could be out, or it could be a goal, or it could be that a foul was committed. 

I propose we change is so that the events `CARRY, RECOVERY, PASS` start a sequence, and the events `BALL_OUT, FOUL_COMMITTED, SHOT` end it. 

What do you think @koenvo?

PR #75, opened 2020-11-11
Improve sequence definition based on latest added generic events

bdagnino:
In this commit, as proposed in #72 I implemented and improvement on the
definition of sequences so that they are opened by a pass, carry or
recovery, and closed by a ball out, fould committed or shot on target.

This results also in sequences in which the team is None, whixh I think
is interesting so you can look at events or tracking data when no team
is in possession.


koenvo:
Can you add some tests to this? Looks like the sequence_open is never set after initialisation and I assume that's not correct

bdagnino:
@koenvo new implementation ready for review! With tests :)

PR #76, opened 2020-11-24
Fix sequences reset at set pieces

bdagnino:
It is difficult to contemplate all the possible ways in which different
providers code for a ball out. This could lead to sequences not being
ended / started properly, leading to a se piece in the middle of a
sequence. In this commit we changed that so that we always start a new
sequence whenever a set piece event occurs.

### DIFF
--- a/kloppy/domain/services/state_builder/builders/sequence.py
+++ b/kloppy/domain/services/state_builder/builders/sequence.py
@@ -31,7 +31,9 @@ class SequenceStateBuilder(StateBuilder):
         return Sequence(sequence_id=0, team=None)
 
     def reduce_before(self, state: Sequence, event: Event) -> Sequence:
-        if isinstance(event, OPEN_SEQUENCE) and state.team != event.team:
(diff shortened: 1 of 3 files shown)


kloppy/domain/services/state_builder/builders/sequence.py
```diff
@@ -31,7 +31,9 @@ def initial_state(self, dataset: EventDataset) -> Sequence:
         return Sequence(sequence_id=0, team=None)
 
     def reduce_before(self, state: Sequence, event: Event) -> Sequence:
-        if isinstance(event, OPEN_SEQUENCE) and state.team != event.team:
+        if isinstance(event, OPEN_SEQUENCE) and (
+            state.team != event.team or event.is_set_piece
```
koenvo, line 35: ‘if event.get_qualifier_type(SetPieceQualifier):’ should also work. Otherwise we might end up with a lot ‘is_....’ methods on event.

Issue #47, opened 2020-08-03
Add length and width property to PitchDimensions


bdagnino:
Currently there is a PitchDiementions object as attribute in the Metadata of a Dataset. The PitchDimension class has attributes `[x,y]_dim`  and`[x,y]_per_meter` but there is no property to directly get the `length` or `width` of the field. 

Would be good to have something like the below implemented so you an get directly get the dimensions of the field in meter without having to do extra calculations. 

```python
@dataclass
class PitchDimensions:
    x_dim: Dimension
    y_dim: Dimension
    x_per_meter: float = None
    y_per_meter: float = None

    @property
    def length(self) -> float:
        return self.x_dim.max / self.x_per_meter

    @property
    def width(self) -> float:
        return self.y_dim.max / self.y_per_meter
```

This would work for the EPTS dataset/metadata, not sure it requires catching exceptions for other providers / formats. Those will need to be taken into account as well.

BenjaminLarrousse:
Is there a place where I can find data spec for providers? 
I imagine _x_per_meter_ and _y_per_meter_ are not always provided. If not, we can always catch the exception and set _length_ and _width_ to None

koenvo:
I think you should not take the data provided into account. Your suggestion to try to return it and when it fails return None sounds very good.
Also take into account that you need `Dimension.mix` in case the min value is other than 0

BenjaminLarrousse:
Sure, should be `self.x_dim.max - self.x_dim.min`

Is there a test file I should put a test for these properties? (I'm still getting used to the package)

koenvo:
Create a new test file please which will contain tests for metadata. test_metadata.py should work

koenvo:
Thank you @BenjaminLarrousse !

BenjaminLarrousse:
> Thank you @BenjaminLarrousse !

You're welcome. I'll get on another issue soon. This one was to for me to start understanding the code, I can do more complicated things now.

PR #77, opened 2020-11-27
Add length and width to PitchDimensions

BenjaminLarrousse:
Add properties to PitchDimensions. Work if x_per_meter and y_per_meter is defined, otherwise return None.
Add a new test file: test_metadata.py, who test the new properties.

Related issue:  #47


kloppy/domain/models/pitch.py
```diff
@@ -21,6 +21,22 @@ class PitchDimensions:
     x_per_meter: float = None
     y_per_meter: float = None
 
+    @property
+    def length(self) -> float:
+        return (
+            (self.x_dim.max - self.x_dim.min) / self.x_per_meter
+            if self.x_per_meter is not None
```
koenvo, line 28: Please remove the ‘is not None’. It should cover both ‘None’ and 0

koenvo:
Not sure why it took me so long to discover the pull request. Looks good. One minor comment in the PR. Thanks for creating this pull request

kloppy/domain/models/pitch.py
BenjaminLarrousse: Sure you're right, I should take into account both instead of one. Do you prefer that it throws and error when it's None or 0, or that both cases are handled ?

koenvo: When you remove “is not None” it’s solved for both cases. No need to throw exceptions

BenjaminLarrousse: Yes sure, but it will throw an error (division by zero).
I'll make the change this week

koenvo: Zero will evaluate to False

BenjaminLarrousse: In fact it fails if I remove the "is not None" condition:

```
   def length(self) -> float:
>       return (self.x_dim.max - self.x_dim.min) / self.x_per_meter
E       TypeError: unsupported operand type(s) for /: 'int' and 'NoneType'
```

I think we should have a condition to be sure it's properly handled.
Maybe I did not understood what you meant ?

koenvo: Ah! I understand the confusion. I was unclear.

What I meant is that both `None` and `0` are falsy ( https://www.freecodecamp.org/news/truthy-and-falsy-values-in-python/ ). So `X if condition else Y` will cover both `None` and `0`

BenjaminLarrousse: Ah! Alright so that it what I was thinking at first: add a condition on 0 and not remove the condition on None.

something like
` if (self.x_per_meter is not None) & (self.x_per_meter != 0)`

koenvo: that can be simplified to `if self.x_per_meter` :-)

BenjaminLarrousse: True. That means I now understand what you were saying at the beginning (and that I should rest)

PR #83, opened 2020-12-23
More docs

koenvo:
This PR contains docs for event part of kloppy. All domain models for events, common and pitch are covered.


docs/api/helpers.md
```diff
@@ -0,0 +1,5 @@
+::: kloppy.domain.services.state_builder.add_state
+
+::: kloppy.helpers.to_pandas
+
+::: kloppy.helpers.transform
```
bdagnino, line 5: @koenvo it seems like helpers.md is not included in _mkdocs.yml_, is that intentional?

bdagnino:
@koenvo reviewed! Great work documenting the API. I quite like how docstrings and mkdocs work together here. Except a comment regarding _helpers.md_ not being included in the documentation looks ready to merge!

docs/api/helpers.md
koenvo: Was intentional but confusing :-)

Changed it a bit

koenvo:
> @koenvo reviewed! Great work documenting the API. I quite like how docstrings and mkdocs work together here. Except a comment regarding _helpers.md_ not being included in the documentation looks ready to merge!

Thanks! The only thing I still want to look into is loading everything from a single module and filter. The filter wasn't working as expected ( https://pawamoy.github.io/pytkdocs/#configuration ); it always showed all members..

Issue #82, opened 2020-12-23
Implement BodyPart qualifier for Statsbomb


koenvo:

BenjaminLarrousse:
BodyPart qualifier exists for Goalkeeper, pass and shot.

We could add it inside _parse_pass() for pass and _parse_shot() for shots (inside the Statsbomb's serializer). A bit like "outcome" is created.

The goalkeeper part is not taken care of, I think ?

koenvo:
@BenjaminLarrousse the goalkeeper is not implemented at all indeed. I would prefer to have BodyPart implemented for each provider first, and then add the GoalKeeper event(s). 

Would you like to implement this for Statsbomb?

BenjaminLarrousse:
Yes I do!

The way I see things:

- in models/event.py, add all possible body part values in the BodyPart class (I can add values for pass, shot, goalkeeper even if we don't use goalkeeper values for now)
- in the Statsbomb's serializer, in _parse_pass and _parse_shot, set a _body_part_ value based on the id inside the source file. For example, `body_part = BodyPart.LEFT_FOOT` if  `pass_dict["body_part"]["id"] = 38`. add a _body_part_ key in the returned dict of the function.
At the top of the file, I can set the value as you did for other events (e.g. SB_BODYPART_LEFT_FOOT = 38)
- Add a test to be sure it works the way it should

Do you see things I should be careful about when implementing this?

koenvo:
That sounds like a great plan @BenjaminLarrousse, and I think you mentioned all important parts 👍

BenjaminLarrousse:
@koenvo  I created a pull request. I think it's good.

Look at it when you have the time and let me know. Thanks

koenvo:
Fixed in https://github.com/PySport/kloppy/pull/86

PR #86, opened 2021-01-20
Add body_part for pass events and shot events in Statsbomb's serializer.

BenjaminLarrousse:
Add body_part feature for pass and shot events:

- Now, the _BodyPart_ class in event has more values: All values are taken from Statsbomb spec
- _PassEvent_ class and _ShotEvent_ class have a new attribute: `body_part`
- __parse_bodypart()_ function is new, and as the name suggests, parse bodypart from source file
- The returned Dict from __parse_pass()_ and __parse_shot()_ has a new key: `body_part`

Statsbomb also has body part for goalkeeper events but they're not yet implemented in kloppy.

Related issue: #82

### DIFF
--- a/kloppy/domain/models/event.py
+++ b/kloppy/domain/models/event.py
@@ -263,14 +263,33 @@ class BodyPart(Enum):
     BodyPart
 
     Attributes:
-        RIGHT_FOOT (BodyPart):
-        LEFT_FOOT (BodyPart):
-        HEAD (BodyPart):
+        RIGHT_FOOT (BodyPart): Pass or Shot with right foot, save with right foot (for goalkeepers).
+        LEFT_FOOT (BodyPart): Pass or Shot with leftt foot, save with left foot (for goalkeepers).
+        HEAD (BodyPart): Pass or Shot with head, save with head (for goalkeepers).
+        BOTH_HANDS (BodyPart): Goalkeeper only. Save with both hands.
+        CHEST (BodyPart): Goalkeeper only. Save with chest.
+        LEFT_HAND (BodyPart): Goalkeeper only. Save with left hand.
+        RIGHT_HAND (BodyPart): Goalkeeper only. Save with right hand.
+        DROP_KICK (BodyPart): Pass is a keeper drop kick.
+        KEEPER_ARM (BodyPart): Pass thrown from keepers hands.
+        OTHER (BodyPart): Other body part (chest, back, etc.), for Pass and Shot.
+        NO_TOUCH (BodyPart): Pass only. A player deliberately let the pass go past him
+                             instead of receiving it to deliver to a teammate behind him.
+                             (Also known as a "dummy").
     """
 
     RIGHT_FOOT = "RIGHT_FOOT"
     LEFT_FOOT = "LEFT_FOOT"
     HEAD = "HEAD"
+
+    BOTH_HANDS = "BOTH_HANDS"
+    CHEST = "CHEST"
+    LEFT_HAND = "LEFT_HAND"
+    RIGHT_HAND = "RIGHT_HAND"
+    DROP_KICK = "DROP_KICK"
+    KEEPER_ARM = "KEEPER_ARM"
+    OTHER = "OTHER"
+    NO_TOUCH = "NO_TOUCH"
 
 
 @dataclass
@@ -383,6 +402,8 @@ class ShotEvent(Event):
         result (ShotResult): See [`ShotResult`][kloppy.domain.models.event.ShotResult]
     """
 
+    body_part: BodyPartQualifier
+
     result: ShotResult
     result_coordinates: Point = None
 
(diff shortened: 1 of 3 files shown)


koenvo:
Thanks for this pull request! 

Please make sure you pass the `BodyPartQualifier` as `qualifiers` attribute to the `Event` class. An example of how qualifiers are parsed within statsbomb serializer can be found here: https://github.com/PySport/kloppy/blob/master/kloppy/infra/serializers/event/statsbomb/serializer.py#L145

It looks like you can move you code to determine body part into that function. Later on the function can also be used for events like GoalKeeper.

BenjaminLarrousse:
AH! I hesitated to do that. As there is only the "type" of event as qualifiers, I thought it was not the right place.
I can move my code to match your request.

Thanks for the feedback.

BenjaminLarrousse:
@koenvo it should be better now.

Let me know if you see anything else.

kloppy/domain/models/event.py
```diff
@@ -389,6 +409,8 @@ class ShotEvent(Event):
     event_type: EventType = EventType.SHOT
     event_name: str = "shot"
 
+    body_part: BodyPartQualifier = None
```
koenvo, line 412: This should be removed

kloppy/domain/models/event.py
```diff
@@ -403,6 +425,7 @@ class PassEvent(Event):
         receiver_coordinates (Point): See [`Point`][kloppy.domain.models.pitch.Point]
         receiver_player (Player): See [`Player`][kloppy.domain.models.common.Player]
         result (PassResult): See [`PassResult`][kloppy.domain.models.event.PassResult]
+        body_part (BodyPartQualifier): body part attributes.
```
koenvo, line 428: This should be removed

kloppy/domain/models/event.py
```diff
@@ -414,6 +437,8 @@ class PassEvent(Event):
     event_type: EventType = EventType.PASS
     event_name: str = "pass"
 
+    body_part: BodyPartQualifier = None
```
koenvo, line 440: This should be removed

kloppy/tests/test_statsbomb.py
```diff
@@ -68,6 +69,14 @@ def test_correct_deserialization(self):
             attacking_direction=AttackingDirection.NOT_SET,
         )
 
+        for qualifier in dataset.events[791].qualifiers:
+            if qualifier == BodyPartQualifier:
+                assert qualifier.value == "HEAD"
+
+        for qualifier in dataset.events[2231].qualifiers:
```
koenvo, line 76: This won’t fail when the qualifier is not set at all.

The get classifier by type might can help you here

kloppy/domain/models/event.py
BenjaminLarrousse: Indeed. Done

kloppy/domain/models/event.py
BenjaminLarrousse: Done

kloppy/domain/models/event.py
BenjaminLarrousse: Done

kloppy/tests/test_statsbomb.py
BenjaminLarrousse: I must admit my test is a bit weak.

Do you mean the get_qualifier_value() function ? If so, I'll make the change

koenvo: Ah Yes that one. That way you can always do the assert. Which will fail in case the qualifier is not set or when it’s set incorrect

BenjaminLarrousse: Allright ! Thanks, I'll commit the changes

kloppy/tests/test_statsbomb.py
```diff
@@ -68,6 +69,20 @@ def test_correct_deserialization(self):
             attacking_direction=AttackingDirection.NOT_SET,
         )
 
+        assert (
+            dataset.events[791].get_qualifier_value(BodyPartQualifier).value
+            == "HEAD"
+        )
+
+        assert (
+            dataset.events[2231].get_qualifier_value(BodyPartQualifier).value
```
koenvo, line 78: Please change this to `dataset.events[2231].get_qualifier_value(BodyPartQualifier) == BodyPart.RIGHT_FOOT`?

koenvo:
About ready to get merged. I added one minor comment about the tests

kloppy/tests/test_statsbomb.py
BenjaminLarrousse: Good idea. It's done

BenjaminLarrousse:
I think we're up and running

koenvo:
Thank you @BenjaminLarrousse !

Issue #53, opened 2020-08-14
Be able to extend / pass your own Event/Frame class


koenvo:
Providing the classes for Events / Frames opens the possibility to add additional:
1. functionality to events / frames
2. attributes to events

Especially 2) is very useful. Providers add very useful information that are not common shared. With just a single implementation of an `Event` we have to throw away this information. 

When we are able to extend the `Event` class, we can pass a provider specific implementation of an Event like `StatsbombShotEvent` which holds more attributes.

PR #87, opened 2021-01-25
Default normalization of coordinates

bdagnino:
[DRAFT]

In this commit I submit the a draft version of the default normalization
of coordiantes for the serializer of Tracab data. I submit this draft so
that we can discuss / modify / agree on an approach before implementing
it on the other serializers.

I have added a VerticalOrietnation model to keep track of the vertical
orientation of the dataset, as it can be reversed when going to original
to default and vice versa. This attribute is at the Metadata level, but
not sure if it should go on the Dimension or PitchDimensions model.

I added length and widht as attributes to the PitchDimensions model,
rather than them being a property, so that they can remain unchanged
when changing the coordinate system of the data.

In the serializer itself, I added the transformation on each frame
depending on the option given to it (normalize by default).

One question I have is whether we should keep a list of models /
coordiantes systems per provider, so that you can transform to a certain
coordiante system, or you can transform back after it being normalized,
for example `transform_to_coordiante_system(CoordianteSystem.TRACAB)`.
And then make Kloppy ber one of them and transform to that one by
default.


koenvo:
This looks good.

One thing I would like to prevent is having a lot of very similar code in different serializers. Therefore it might be good to introduce a `DatasetBuilder` which contains all logic to go from dicts to Point and Frame.

Some advantages:
1. Keep all logic to normalize coordinates at one place
2. Open up the opportunity to introduce more transformations during load time
3. Possibility to load directly into other formats than the domain models

Disadvantages:
1. Another layer
2. Introduce an intermediate format to pass data from Serializer to DatasetBuilder (this will probably be just dictionaries), which can lead to performance penalty

Furthermore the length and width should not be changed as the `x_per_meter` and `y_per_meter` changed during normalization, right?

bdagnino:
@koenvo  True about the length and width! 

About the DatasetBuilder, I'm not sure I get what you mean. I 100% agree it would be best not to have so much code pseudo-repeated in every serialized, but at the same time the format of the dicts in each provider is quite different, so we might end up with pseudo-repeated code somewhere else. 

But most likely I'm not getting it right. Can you elaborate may be with some pseudo-code to guide me?

koenvo:
Something like this:

```python
class TrackingDatasetBuilder:
  def __init__(self, options):
    self.should_normalize = options.get('normalize', False)
    self.reset()

  def reset():
    self.frames = []
    self.metadata = None

   def set_metadat(metadata: Metadata):
     self.metadata = metadata

  # Only pass build-in types ( https://docs.python.org/3/library/stdtypes.html )
  def add_frame(self, timestamp: float, ball_position: Tuple[float, float], player_positions: Iterator[Tuple[str, Tuple[float, float]]], ...):
     assert self.metadata is not None, "call set_metadata first"
    if self.should_normalize:
     ....
    frame = Frame(
      ...
    )
    self.frames.append(frame)

class Serializer:
  def __init__(self, dataset_builder):
    self.dataset_builder = dataset_builder

  def deserializer(...):
    self.dataset_builder.set_metadata(...)

    for line in data:
       self.dataset_builder.add_frame(
           float(line[0]),
           (
               (player_info[0], (float(player_info[1]), float(player_info[2])))
               for player_info in
               (cell.split(',') for cell in line[1].split(';'))
           )
    
```

This would split the code in two components:
1. Deserialization
2. Dataset construction

In between should only built-in types be used.

This design would also make implementation of https://github.com/PySport/kloppy/issues/53 super easy, as construction of domain objects (Frame and Event) will happen only at one place. It would help us also to get rid of the `Event.create` calls.

Curious what you think about it.

bdagnino:
Hi @koenvo!

I'm a bit stuck with this. I tried to implement what you described above but couldn't really figure out how to implement it properly. It seems to me that this is more something to be done as a refactor of the way serializers are implemented than something to be done as part of an issue to set a default normalization for all providers. I think d rather focus on the normalization itself at the moment, and we can refactor the way things are structured later. What do you think?

On the specific point f the normalization I think it could be interesting to set it up with model for all possible coordinates systems, and then methods that take individual frames, individual events, metadata or whole events or tracking datasets from one system to the other. 

This is more versatile because you can change data to whatever coordinate system you prefer, even data from different providers. and the normalization would be just one case of all those possible cases. What do you think?

koenvo:
Hmm yes good point. Lets keep the refactor for later. When the standard coordinates are implemented more details about how a (potential) refactor would look like will appear. 

Are you able to go forward without the refactor?

bdagnino:
@koenvo what do you think would be the best way to create the models for the different providers coordinantes system? Initially I was thinking of extending the `Providers` Enum to have the coordinates system information of each provider, but I think you can't do that with Enums? 

So another option is to change the Providers from an Enum to a dataclass, or another option, to create a different dataclass just for coordinates systems? That would allow to have one attribute in the metadata for Provider, and another one for CoordinateSystem.

I think I favor the later, what do you think?

kloppy/domain/models/common.py
```diff
@@ -294,6 +302,79 @@ def duration(self):
...
+    width: float = None
+
+    @property
+    def pitch_dimensions(self) -> PitchDimensions:
+        return PitchDimensions(
+            x_dim=Dimension(-1 * self.length / 2, self.length / 2),
+            y_dim=Dimension(-1 * self.width / 2, self.width / 2),
+            x_per_meter=10000,
+            y_per_meter=10000,
+        )
```
bdagnino, line 375: @koenvo Is there a better way to do this than by defining dataclasses? Ideally I'd like something like an Enum with properties, but not sure that is something possible.

kloppy/domain/models/common.py
```diff
@@ -294,6 +302,79 @@ def duration(self):
...
+    CENTER = "center"
+
+    def __str__(self):
+        return self.value
+
+
+@dataclass
+class CoordinateSystem(ABC):
+    provider: Provider
+    origin: Origin
```
bdagnino, line 324: @koenvo The `origin` attribute is not actually needed because it's implicit on the `dimension`, but I think is better to keep it because it's transparent, descriptive and could be useful down the line.

bdagnino:
@koenvo What do you think of this new implementation? 

I'm quite happy with it tbh 😝. I like that there is one place where all the coordinate systems are described, and that the changes to the serializer are actually minimal, because all the logic is inside the transformer, and it's not replicated on each serializer. 

I also like how the `CoordianteSystem` have a `PitchDimension` property, and also that all the previous transformer APIs are kept in place. 

What I don't like is that the `CoordinateSystems` are dataclasses and that you have to instantiate an instance to pass to the transformers, but not sure there is a better way to do it. 

There are many things to polish, specially inside the serializer and how the default case is handle, and need to create an equal dunder to skip the transformation if the origin and destine coordinate systems are the same, etc.

However wanted to get your feedback on this draft implementation to see if we agree this is the right way to do it to then apply it on other serializers and coordinate systems as well. 

What do you think?

kloppy/domain/models/common.py
koenvo: Well.. you could just pass a class around instead of an instance of it. That would work in some cases. But in this case you also have instance attributes (length, width) which are tied to a single instance of `TracabCoordinateSystem`.

This information is important as you can transform from TRACAB 105x68 to TRACAB 108x70 (not sure if dimensions make sense). If you use classes instead of instances it will act like a singleton and when you change attributes on one, the attributes will also be changed on the other (reference).

You could - at the bottom of the code for example - initialise some most-used versions.

```python
defaults = {
  Provider.TRACAB: TracabCoordinateSystem(normalized=False, length=105, ...)
}
```

kloppy/domain/models/common.py
koenvo: Totally agree

kloppy/domain/models/common.py
```diff
@@ -294,6 +302,79 @@ def duration(self):
...
+    vertical_orientation: VerticalOrientation
+    normalized: bool
```
koenvo, line 326: The `provider`, `origin` and `vertical_orientation` are attributes that are tied to a single - valid - value per class. 

Consider using `abstractmethod` and `property` decorators like here: https://github.com/PySport/kloppy/blob/master/kloppy/domain/models/event.py#L343-L346
This way you can enforce that these properties are defined, but prevent them from being passed with a potential wrong value. You can achieve similar behaviour by using `__post__init__` ( https://docs.python.org/3/library/dataclasses.html#post-init-processing ) but I believe in that case that doesn't make sense as there is only 1 single value that is valid.

kloppy/domain/services/transformers/__init__.py
```diff
@@ -74,11 +92,45 @@ def __needs_flip(
         return flip
 
     def transform_frame(self, frame: Frame) -> Frame:
-        flip = self.__needs_flip(
+
+        # Change coordinate system
+        if self._to_coordinate_system is not None:
```
koenvo, line 97: I really like you did this in 3 separate steps. I might come with a performance penalty but can be solved by saving the transformed output before continue processing, so you don't have to do the transformation everytime

kloppy/infra/serializers/tracking/tracab.py
```diff
@@ -154,6 +158,9 @@ def deserialize(
         sample_rate = float(options.get("sample_rate", 1.0))
         limit = int(options.get("limit", 0))
         only_alive = bool(options.get("only_alive", True))
+        to_coordinate_system = options.get(
```
koenvo, line 161: Aha.. now I see what you mean. I would pass a class instead of an instance.
`to_coordinate_system=KloppyCoordinateSystem`

bdagnino: @koenvo I like the idea of passing a name! May be even better if the builder takes in the `Provider` Enum?

koenvo: Uhm yes. Accept the exact value of the `Provider` enum. But lets not bother the end-user with passing the `Provider.TRACAB` value itself.

kloppy/domain/models/common.py
```diff
@@ -294,6 +304,240 @@ def duration(self):
...
+    def provider(self) -> Provider:
+        return Provider.SPORTEC
+
+    @property
+    def origin(self) -> Origin:
+        return Origin.BOTTOM_LEFT
+
+    @property
+    def vertical_orientation(self) -> VerticalOrientation:
+        return VerticalOrientation.TOP_TO_BOTTOM
```
bdagnino, line 458: @koenvo I don't know what is the correct coordinate system for Sportec. Can you provide that info?

bdagnino: @koenvo just a reminder that I need this info!

bdagnino:
@koenvo this PR is ready for review!

Besides the feedback you might provide, there are 2 things that might need changes: 

1. The CoordinateSystem for Sportec might not be accurate. 

2. I think we need to look into whether we want to keep the `pitch_dimensions` attribute in the `Metadata` together with the `coordinate_sytem` attribute, or instead only keep the `coordiante_system` one (this is what I lean towards to). Having both is problematic in a couple of ways. For example you can't do both on the same transformation and need to handle that, but also you could have one pitch dimensions in the `metadata`, and another one in `metadata.coordinate_system`.  Given that now pitch dimensions are included in the coordinate system I don't see a reason to keep them separate. What do you think?

Other than that, it's ready for review! Let me know what do you think!

bdagnino:
@koenvo I added the normalization for skillcorner serializer, so this PR is ready for a final review before working on the documentation. 

I haven't changed anything regarding the Z dimension, so the values in the Z coordinate will be always left the same as they come. That will need to change probably in the future, but we can leave that for another PR. It could be a good beginners issue may be?

kloppy/domain/models/pitch.py
```diff
@@ -32,34 +39,14 @@ class PitchDimensions:
 
     x_dim: Dimension
     y_dim: Dimension
-    x_per_meter: float = None
-    y_per_meter: float = None
+    length: float = None
+    width: float = None
 
-    @classmethod
```
bdagnino, line 38: @koenvo I deleted this default method, and also set the serializer for XML to have None as pitch dimensions. 

I think if there are no coordinate system or pitch dimensions is better to not assume anything / not have any default. What do you think?

kloppy/infra/serializers/code/xml.py
```diff
@@ -63,12 +63,13 @@ def deserialize(
             metadata=Metadata(
                 teams=[],
                 periods=[period],
-                pitch_dimensions=PitchDimensions.default(),
```
bdagnino, line 66: See my other comment on this topic.

kloppy/domain/services/transformers/__init__.py
```diff
@@ -11,41 +11,86 @@
...
+        from_coordinate_system: CoordinateSystem = None,
+        from_pitch_dimensions: PitchDimensions = None,
+        from_orientation: Orientation = None,
+        to_coordinate_system: CoordinateSystem = None,
+        to_pitch_dimensions: PitchDimensions = None,
+        to_orientation: Orientation = None,
     ):
-        self._from_pitch_dimensions = from_pitch_dimensions
+
+        print("hello")
```
koenvo, line 34: Guess this shouldn't be here :-)

kloppy/domain/services/transformers/__init__.py
```diff
@@ -87,46 +174,139 @@ def transform_frame(self, frame: Frame) -> Frame:
...
             players_coordinates={
-                key: self.transform_point(point, flip)
+                key: self.change_point_dimensions(point)
+                for key, point in frame.players_coordinates.items()
+            },
+        )
+
+    def __change_point_coordinate_system(self, point: Union[Point, Point3D]):
+
+        if point is None:
```
koenvo, line 188: `if point:`

kloppy/domain/services/transformers/__init__.py
```diff
@@ -11,41 +11,86 @@
...
-            y=self._to_pitch_dimensions.y_dim.from_base(y_base),
-        )
+        if isinstance(point, Point3D):
+            return Point3D(x=x, y=y, z=point.z)
+        else:
+            return Point(x=x, y=y)
+
+    def flip_point(self, point: Union[Point, Point3D]):
+
+        if point is None:
```
koenvo, line 78: `if not point:`

kloppy/domain/services/transformers/__init__.py
```diff
@@ -11,41 +11,86 @@
...
+        from_orientation: Orientation = None,
+        to_coordinate_system: CoordinateSystem = None,
+        to_pitch_dimensions: PitchDimensions = None,
+        to_orientation: Orientation = None,
     ):
-        self._from_pitch_dimensions = from_pitch_dimensions
+
+        print("hello")
+        if (
+            from_pitch_dimensions is not None
```
koenvo, line 36: `if from_pitch_dimensions and from_coordinate_system [etc]`

those variables shouldn't have any falsy value, other than `None`

kloppy/domain/services/transformers/__init__.py
```diff
@@ -74,11 +119,53 @@ def __needs_flip(
...
+        # Flip frame based on orientation
+        if self.__needs_flip(
             ball_owning_team=frame.ball_owning_team,
             attacking_direction=frame.period.attacking_direction,
+        ):
+            frame = self.__flip_frame(frame)
+
+        return frame
+
+    def __needs_coordinate_system_change(self):
```
koenvo, line 139: please set as property in the `__init__`

kloppy/domain/services/transformers/__init__.py
```diff
@@ -74,11 +119,53 @@ def __needs_flip(
...
+        return self._from_coordinate_system != self._to_coordinate_system
+
+    def __needs_pitch_dimensions_change(self):
```
koenvo, line 142: please set as property in the `__init__`

kloppy/domain/services/transformers/__init__.py
```diff
@@ -87,46 +174,139 @@ def transform_frame(self, frame: Frame) -> Frame:
...
+        dataset: Dataset,
         to_pitch_dimensions: PitchDimensions = None,
         to_orientation: Orientation = None,
-    ) -> DatasetT:
-        if not to_pitch_dimensions and not to_orientation:
+        to_coordinate_system: CoordinateSystem = None,
+    ) -> Dataset:
+
+        if (
+            to_pitch_dimensions is not None
```
koenvo, line 295: `if to_pitch_dimensions and to_coordinate_system`

kloppy/domain/services/transformers/__init__.py
```diff
@@ -87,46 +174,139 @@ def transform_frame(self, frame: Frame) -> Frame:
...
+            and to_coordinate_system is not None
+        ):
+            raise ValueError(
+                "You can't do both a PitchDimension and CoordinateSysetm on the same dataset transformation"
+            )
+
+        if (
```
koenvo, line 302: great!

kloppy/domain/services/transformers/__init__.py
```diff
@@ -135,12 +315,33 @@ def transform_dataset(
                     "ball owning team data"
                 )
 
-        transformer = cls(
-            from_pitch_dimensions=dataset.metadata.pitch_dimensions,
-            from_orientation=dataset.metadata.orientation,
-            to_pitch_dimensions=to_pitch_dimensions,
-            to_orientation=to_orientation,
-        )
+        if to_pitch_dimensions is not None:
```
koenvo, line 318: `if to_pitch_dimensions`

kloppy/domain/services/transformers/__init__.py
```diff
@@ -135,12 +315,33 @@ def transform_dataset(
...
+
+            transformer = cls(
+                from_pitch_dimensions=dataset.metadata.pitch_dimensions,
+                from_orientation=dataset.metadata.orientation,
+                to_pitch_dimensions=to_pitch_dimensions,
+                to_orientation=to_orientation,
+            )
+
+        elif to_coordinate_system is not None:
```
koenvo, line 327: `elif to_coordinate_system`

kloppy/helpers.py
```diff
@@ -229,11 +241,24 @@ def transform(
...
-        to_orientation=to_orientation,
-        to_pitch_dimensions=to_pitch_dimensions,
-    )
+        return Transformer.transform_dataset(
+            dataset=dataset,
+            to_orientation=to_orientation,
+            to_pitch_dimensions=to_pitch_dimensions,
+        )
+
+    if to_coordinate_system and (isinstance(to_coordinate_system, Provider)):
```
koenvo, line 250: redundant parentheses

kloppy/helpers.py
```diff
@@ -217,10 +219,20 @@ def write_xml_code_data(dataset: CodeDataset, xml_filename: str):
...
 def transform(
-    dataset: DatasetT, to_orientation=None, to_pitch_dimensions=None
-) -> DatasetT:
+    dataset: Dataset,
+    to_orientation=None,
+    to_pitch_dimensions=None,
+    to_coordinate_system=None,
+) -> Dataset:
+
+    if to_pitch_dimensions is not None and to_coordinate_system is not None:
```
koenvo, line 228: `if to_pitch_dimensions and to_coordinate_system`

kloppy/helpers.py
```diff
@@ -217,10 +219,20 @@ def write_xml_code_data(dataset: CodeDataset, xml_filename: str):
 
 
 def transform(
-    dataset: DatasetT, to_orientation=None, to_pitch_dimensions=None
-) -> DatasetT:
+    dataset: Dataset,
+    to_orientation=None,
+    to_pitch_dimensions=None,
+    to_coordinate_system=None,
```
koenvo, line 225: The `to_coordinate_system` does also accept `Provider`. Think it's better to change the signature of the function and specify the argument types to help us remind what types are accepted.

kloppy/infra/serializers/event/wyscout/serializer.py
```diff
@@ -275,6 +277,23 @@ def deserialize(
         if not options:
             options = {}
 
+        from_coordinate_system = build_coordinate_system(
+            Provider.SPORTEC,
```
koenvo, line 281: This provider seems to be incorrect

kloppy/infra/serializers/tracking/epts/metadata.py
```diff
@@ -266,6 +267,15 @@ def load_metadata(
 
     metadata.orientation = orientation
 
+    if provider is not None and pitch_dimensions is not None:
```
koenvo, line 270: `if provider and pitch_dimensions`

kloppy/infra/serializers/tracking/epts/serializer.py
```diff
@@ -59,6 +65,11 @@ def _frame_from_row(row: dict, metadata: EPTSMetadata) -> Frame:
             ),
         )
 
+        if transformer is not None:
```
koenvo, line 68: `if transformer`

kloppy/infra/serializers/tracking/epts/serializer.py
```diff
@@ -112,10 +123,27 @@ def deserialize(
         with performance_logging("Loading metadata", logger=logger):
             metadata = load_metadata(inputs["metadata"])
 
+            if (
+                metadata.provider is not None
```
koenvo, line 127: `if metadata.provider and metadaa.pitch_dimensions`

kloppy/infra/serializers/tracking/epts/serializer.py
```diff
@@ -127,6 +155,13 @@ def deserialize(
                 )
             ]
 
+        if transformer is not None:
```
koenvo, line 158: `if transformer`

kloppy/tests/test_opta.py
```diff
@@ -59,3 +61,20 @@ def test_correct_deserialization(self):
...
+
+        with open(f"{base_dir}/files/opta_f24.xml", "rb") as f24_data, open(
+            f"{base_dir}/files/opta_f7.xml", "rb"
+        ) as f7_data:
+
+            dataset = serializer.deserialize(
+                inputs={"f24_data": f24_data, "f7_data": f7_data}
+            )
+
+        assert dataset.events[0].coordinates == Point(0.501, 0.506)
```
koenvo, line 80: Opta is 100x100. That does mean normalised to 1x1 should result in `0.501, 0.494`, right?

kloppy/domain/models/common.py
```diff
@@ -303,6 +313,261 @@ def __eq__(self, other):
...
+    def provider(self) -> Provider:
+        return Provider.KLOPPY
+
+    @property
+    def origin(self) -> Origin:
+        return Origin.TOP_LEFT
+
+    @property
+    def vertical_orientation(self) -> VerticalOrientation:
+        return VerticalOrientation.TOP_TO_BOTTOM
```
koenvo, line 381: I'm not entirely sure we should default to `TOP_TO_BOTTOM`. I personally prefer the Cartesian coordinate system

kloppy/tests/test_opta.py
koenvo: oh. see my comment about `vertical_orientation` of KloppyCoordinateSystem

bdagnino:
@koenvo All comments addressed! I also added some basic documentation regarding the normalization of coordinates and how to either take advantage of it, or preserve the original coordinate system. 

We could probably use better documentation on that regard, but I would say this is enough to do a first release. What do you think?

Hopefully this is ready to be merged!

PR #90, opened 2021-03-31
Skillcornertrackingserializer

koenvo:


kloppy/infra/datasets/tracking/skillcorner.py
```diff
@@ -0,0 +1,32 @@
...
+from io import BytesIO
+
+from ..core.builder import DatasetBuilder
+from ...serializers.tracking import (
+    TrackingDataSerializer,
+    SkillCornerTrackingSerializer,
+)
+
+
+with urllib.request.urlopen('https://raw.githubusercontent.com/SkillCorner/opendata/master/data/matches.json') as url:
```
koenvo, line 12: This will cause a request everytime you load kloppy. Please move this logic into `SkillCornerTracking`, and cache it within the class.

But as you can see in other dataset loads we don't really validate the input and just send a request to the source. Might be cool to be able to list all available dataset for a source.

kloppy/infra/serializers/tracking/skillcorner.py
```diff
@@ -0,0 +1,427 @@
...
+                    team = home_team
+                elif team_id == away_team_id:
+                    team_string = "AWAY"
+                    team = away_team
+
+                players[team_string][player_id] = Player(
+                    player_id=f"{team.ground}_{player['number']}",
+                    team=team,
+                    jersey_no=player["number"],
+                    name=" ".join([player["first_name"], player["last_name"]]),
```
koenvo, line 343: Please use f-strings here: 
```python
f"{player['first_name']} {player['last_name']}"
```

kloppy/infra/serializers/tracking/skillcorner.py
```diff
@@ -0,0 +1,427 @@
...
+                for frame in raw_data:
+                    frame_period = frame["period"]
+
+                    if frame_period is not None:
+                        if n % sample == 0:
+                            yield frame
+                        n += 1
+
+        frames = []
+        for n, _frame in enumerate(_iter()):
```
koenvo, line 375: Did you consider moving the logic from within `_iter` to within this loop?

kloppy/infra/datasets/tracking/skillcorner.py
znstrider: I think I would just remove this logic completely and replace it with

`_match_ids = [4039, 3749, 3518, 3442, 2841, 2440, 2417, 2269, 2068]`

In the same way I could list the match info strings and make it available through a function
get_available_matches within the SkillCornerTracking(DatasetBuilder) which returns the list below.
Does that sound good?

```
_match_info_strings =
['ID: 4039 - Manchester City vs Liverpool on 2020-07-02',
 'ID: 3749 - Dortmund vs Bayern Munchen on 2020-05-26',
 'ID: 3518 - Juventus vs Inter on 2020-03-08',
 'ID: 3442 - Real Madrid vs FC Barcelona on 2020-03-01',
 'ID: 2841 - FC Barcelona vs Real Madrid on 2019-12-18',
 'ID: 2440 - Liverpool vs Manchester City on 2019-11-10',
 'ID: 2417 - Bayern Munchen vs Dortmund on 2019-11-09',
 'ID: 2269 - Paris vs Marseille on 2019-10-27',
 'ID: 2068 - Inter vs Juventus on 2019-10-06']


class SkillCornerTracking(DatasetBuilder):
    def get_available_matches(self) -> List[str]:
        return _match_info_strings
```

kloppy/infra/serializers/tracking/skillcorner.py
znstrider: What's the advantage of doing so?

I oriented this after the TRACABSerializer, which does it similarly.

kloppy/infra/datasets/tracking/skillcorner.py
koenvo: I like the idea to make a list available. For statsbomb that's a huge list. And I guess they don't have a single list with all matches in it. Do you think there is a way to solve this? 

Otherwise we can move this new feature to a new issue and discuss it there, and leave it out for this PR.

kloppy/infra/serializers/tracking/skillcorner.py
koenvo: Ah I missed the sample rate code here. Yes. Let's keep it your way.

kloppy/infra/datasets/tracking/skillcorner.py
znstrider: Yeah that's a good point about the consistency between datasets. Best might be to call the api / read in the match lists
as you first proposed within each DatasetBuilder class which could be discussed in a seperate issue. But it's not necessary I supoose.

It would be great to have some sort of list like above available somewhere though - that data for top level matches is kinda special.
Could we add the list along with a short introduction here: https://kloppy.pysport.org/quickstart/tracking_data/ or is a seperate quickstart page for broadcast tracking data better?

koenvo: This looks great @znstrider!  Really like to additional notebook file!

I think the 50MB file is a bit big. But i'll make sure it's not included in the final release and I might cut it down a bit some day :-) Shall I also add the datasets code (after it's merged)?

Issue #92, opened 2021-04-06
Serializer for sportscode like XML


bdagnino:
Some clubs get data from instat or wyscout which comes in form of "codes" XML format. This format is the "sportscode xml" format, meaning is how sportscode export the coding of a game in XML format. Is not very structured data, nothing like event type data, but more like coding of phases of the game. Build up by this team x seconds to y seconds, counter by the opponent form x seconds to y seconds, etc. Specially colleges in the US get data in this format, and they want to use it to then know for examples what happens on their defensive phases and that type of things. 

A typical bit of xml file looks like this. 

```xml
<file>
 <SORT_INFO>
  <sort_type>sort order</sort_type>
 </SORT_INFO>
 <ALL_INSTANCES>
  <instance>
   <ID>0</ID>
   <code>Offsets</code>
   <start>2</start>
   <end>5</end>
   <label>
    <text>First half start</text>
    <group>Offset</group>
   </label>
  </instance>
  <instance>
   <ID>4</ID>
   <code>Duke Blue Devils - Opposition build-up play</code>
   <start>2</start>
   <end>13</end>
   <label>
    <group>Event</group>
    <text>Opposition build-up play</text>
   </label>
  </instance>
 </ALL_INSTANCES>
 <ROWS>
  <row>
   <sort_order>1</sort_order>
   <code>Offsets</code>
   <R>0</R>
   <G>0</G>
   <B>0</B>
  </row>
  <row>
   <sort_order>2</sort_order>
   <code>Duke Blue Devils - Opposition build-up play</code>
   <R>65535</R>
   <G>65535</G>
   <B>65535</B>
  </row>
 </ROWS>
</file>
```
Since there are no unique code identifiers and there is no guaranteed metadata on each file (not specific provider), everyone can code or include any information they want, I'd suggest not to add them as another event serializer, but rather include it as another serializer type. 

May be have serializer:
events
tracking
codes

And have the xml serializer return a CodesDataset that doesn't have any metadata, and just the records, with a method to go from the records to a dataframe?

@koenvo what do you think?

PR #96, opened 2021-05-15
Xml serializer

koenvo:
- [x] Tests
- [x] XMLCodeSerializer for both reading and writing
- [x] Update kloppy-query
- [x] Update docs
- [x] to pandas


bdagnino:
@koenvo looks great! Only slightly disappointed that I won't get the beers! :P

I have only one comment. Usually on the XMLs that are sportscode like, you have the codes defined at the end of the file (with the name of the code, the order and the color). See [the example  in the original issue](https://github.com/PySport/kloppy/issues/92). I see that in the serializer and on the tests files that is not contemplated / included. 

Ideally would be good to keep the color information stored, but for now it's fine as it is. I only mention it to make sure that the serializer doesn't break when reading a file with the codes information at the end. Also not sure what happens if you try to import a file on sports code or other software if there is no code information at the end.

koenvo:
> @koenvo looks great! Only slightly disappointed that I won't get the beers! :P
> 
> I have only one comment. Usually on the XMLs that are sportscode like, you have the codes defined at the end of the file (with the name of the code, the order and the color). See [the example in the original issue](https://github.com/PySport/kloppy/issues/92). I see that in the serializer and on the tests files that is not contemplated / included.
> 
> Ideally would be good to keep the color information stored, but for now it's fine as it is. I only mention it to make sure that the serializer doesn't break when reading a file with the codes information at the end. Also not sure what happens if you try to import a file on sports code or other software if there is no code information at the end.

I'll sent the beer anyway :-)

I was not able to verify that the 'sort_order' is optional but I did generate XML files before and I didn't get complains about it.. can also mean they weren't used. But I assume the current format in this PR works.

Can you provide an example of a workflow when the colours are useful?

PR #98, opened 2021-05-25
add ball_z to skillcornertrackingserializer

znstrider:
Adds the ball_z to ball_coordinates and includes the corresponding tests.

Also adds the ball_z into the _frame_to_pandas_row_converter function in helpers.py

### DIFF
--- a/kloppy/helpers.py
+++ b/kloppy/helpers.py
@@ -230,7 +230,6 @@ def _frame_to_pandas_row_converter(frame: Frame) -> Dict:
         else None,
         ball_x=frame.ball_coordinates.x if frame.ball_coordinates else None,
         ball_y=frame.ball_coordinates.y if frame.ball_coordinates else None,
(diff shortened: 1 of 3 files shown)


znstrider:
@koenvo
What do you think about this?
At the moment to_pandas doesn't return the z_coordinate.

If we integrate the z_coordinate into to_pandas, I think we have two options:
1) Changing all providers ball_coordinates to use Point3D even when they don't offer the z_coord.
2) Or we might handle that in to_pandas (eg. by checking for Point vs Point3D in ball_coordinates).
If we do 1) we might still want to check if all ball z_coords are None to not return a complete None column with to_pandas.

Or we might just leave out the z_coord in to_pandas, but to me that seems wrong.

koenvo:
I prefer the second option. Would be good to also update https://github.com/PySport/kloppy/blob/master/kloppy/domain/models/tracking.py#L14

Should the z coordinate be None or 0 if not provided? Guess None so users can see it’s undefined.

Issue #97, opened 2021-05-25
Include all Qualifiers in Metrica Event Serializer 


cedrickrause:
Right now the metrica event json serializer only adds SetPieceQualifier to the kloppy event. BodyPartQualifier and PassQualifier are not included. 

For the BodyPartQualifier the way I see it, the data only includes the subtype "HEADER" if it is a header, but no information about which foot was used for example. 

PassQualifiers that are used can be found in the [documentation of metrica events](https://github.com/metrica-sports/sample-data/blob/master/documentation/events-definitions.pdf). Here it is maybe noteworthy that Goal Kick is considered a pass type in metrica, while it is a set piece in kloppy. Goal Kick is already included as a SetPieceQualifier in the serializer already.

I assume additions need to be made in [_get_event_qualifiers()](https://github.com/PySport/kloppy/blob/3a25d18b38da8a73f01f66b8bc7bd3f12ca94165/kloppy/infra/serializers/event/metrica/json_serializer.py#L149)

cedrickrause:
Is there any further explanation on the [kloppy pass types](https://kloppy.pysport.org/api/domain/event/#kloppy.domain.models.event.PassType)?

The types are not the same as providers', so I am not sure how I would map them.

cedrickrause:
The body part is only given for shots in sportec event data, not for passes. I think it still makes sense to integrate it for the shots.

cedrickrause:
Just to clarify the state of this, since part of it was already merged. 

The Pass Qualifiers are not yet included for the providers mentioned in the original post. BodyPart qualifiers are done. 
Not sure when/if I will do it, but it shouldn't be hard after knowing the mapping between kloppy pass types and the pass types of the providers.

koenvo:
I've updated the title of the issue. Thanks for adding the BodyPart qualifiers!

koenvo:
**Implementation tips for the PassType**

1. Parse the input file and create a list of qualifiers for each event. A pass qualifier is defined like this: `PassQualifier(PassType.CROSS)`. Both can be imported from `kloppy.domain`: `from kloppy.domain import PassQualifier, PassType`. `PassType` is an enum define here https://github.com/PySport/kloppy/blob/8ad513532518b97fc0cea90300f668b38c78d5b5/kloppy/domain/models/event.py#L233-L244 

[...]

PR #99, opened 2021-05-29
Include all event qualifiers

cedrickrause:
DRAFT

Pull Request for adjustments mentioned in #97.
Started with BodyPart for metrica event data.


cedrickrause:
BodyPart for sportec event data also added now. However the body part is only given for shots in sportec data, so there are no body part qualifiers for passes or so.

koenvo:
Thanks for the pull request @cedrickrause! I'm a bit busy. Will have a look at the code and your questions and respond asap

cedrickrause:
@koenvo No worries. 
I now also added the body part qualifier for the opta event data. So all missing serializers should now have the body part qualifier. However there is none of the body part qualifiers in the example opta event f24 file, so finding a new one with them in it, would be needed to test the new behavior.
Might make sense to seperate the pass qualifiers in another pull request, as I am not sure how quickly I will get to it.

koenvo:
Thanks Cedrick!

PR #100, opened 2021-06-24
Modify etps serializer so that it accepts files with no ball z

bdagnino:
I also added files and test to test this. As this is how Metrica Sports
provides the data (no z coordinate) I included this files and test as
testing the serialization of Metrica Sports in particular, which I
believe is something good to have (test for each provider, not just for
ETPS in general)


bdagnino:
@koenvo I spotted this issue while working in codeball. I need to use the latest changes of kloppy master there, but at the moment can't because I can't load the data. Any chance we can make a quick merge of this and release a new version?

PR #101, opened 2021-06-30
Changes to accept non int ProviderParameters and to read all sensors …

bdagnino:
…in EPTS data

### DIFF
--- a/kloppy/infra/serializers/tracking/epts/metadata.py
+++ b/kloppy/infra/serializers/tracking/epts/metadata.py
@@ -40,7 +40,7 @@ def _load_provider_parameters(parent_elm, value_mapper=None) -> Dict:
 
 def _load_periods(global_config_elm, frame_rate: int) -> List[Period]:
     provider_params = _load_provider_parameters(
-        global_config_elm.find("ProviderGlobalParameters"), value_mapper=int
+        global_config_elm.find("ProviderGlobalParameters")
     )
 
     period_names = [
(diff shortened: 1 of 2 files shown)


kloppy/infra/serializers/tracking/epts/metadata.py
```diff
@@ -40,7 +40,7 @@ def _load_provider_parameters(parent_elm, value_mapper=None) -> Dict:
 
 def _load_periods(global_config_elm, frame_rate: int) -> List[Period]:
     provider_params = _load_provider_parameters(
-        global_config_elm.find("ProviderGlobalParameters"), value_mapper=int
```
bdagnino, line 43: This was not needed as the result is the same whether you map to int or not, and it made the serialization crashed when there where provider parameters present that where not numbers.

kloppy/infra/serializers/tracking/epts/serializer.py
```diff
@@ -120,8 +120,8 @@ def deserialize(
                     raw_data=inputs["raw_data"],
                     metadata=metadata,
                     sensor_ids=[
-                        "position"
-                    ],  # we don't care about other sensors
+                        sensor.sensor_id for sensor in metadata.sensors
```
bdagnino, line 123: I believe we should load all sensors in the data, and the user can later filter out the ones they don't want.

koenvo: That sounds like a good addition! The data from the other sensors is not loaded into the frames: https://github.com/PySport/kloppy/blob/master/kloppy/infra/serializers/tracking/epts/serializer.py#L42-L59

We would need a way to store additional information per frame.

kloppy/infra/serializers/tracking/epts/metadata.py
koenvo: Make sense!

PR #111, opened 2021-10-09
Update package description in README

probberechts:
I've created a separate PR for this one since it changes the content of the README. I had the feeling that the initial description was a bit vague. What are (de)serializers, standardized data models, filters and transformers exactly? That might not be clear to many people. I thought it would be better to focus on the problem that Kloppy attempts to solve, rather than the functionality that it provides.


koenvo:
I had the feedback from @willmcgugan to focus more on use-cases instead of just functionality. This is exactly what you did. Thanks a lot for this PR!

ps. I can close the other PR? It seems be a subset of this PR

PR #118, opened 2021-11-18
Simpler api

koenvo:


kloppy/infra/serializers/event/opta/deserializer.py
```diff
@@ -370,89 +368,22 @@ def _get_event_type_name(type_id: int) -> str:
...
-        self.__validate_inputs(inputs)
-        if not options:
-            options = {}
-
-        from_coordinate_system = build_coordinate_system(
-            Provider.OPTA,
-            length=100,
-            width=100,
-        )
+OptaInputs = NamedTuple(
```
willmcgugan, line 371: You can use a class syntax to define NamedTuples

```python
class OptaInputs:
    f7_data: IO[bytes]
    f24_data: IO[bytes]
```

kloppy/infra/serializers/event/sportec/deserializer.py
```diff
@@ -539,27 +511,24 @@ def deserialize(
 
         events = list(
             filter(
-                lambda _event: _include_event(_event, wanted_event_types),
+                lambda _event: self.should_include_event(_event),
```
willmcgugan, line 514: You could replace that lambda with just `self.should_include_event`

willmcgugan (approved):
So I understand the bulk of the change is so that you can do something like:

```python
from kloppy import metrica
data = metrica.load_event("data")
```

Which is super readable. 👍 

Typing is off to a great start. There are some instances of containers with the type unspecified. i.e. plain `Dict`, and not `Dict[str, Foo]`.  It's best to specify these where you can. You can also use a TypedDict if you have a dict with keys of various types.

Typing an ongoing process though. Doesn't have to hold anything up. Mypy has a switch to generate an HTML report. You can use this to see which lines might need more typing information.

I would strongly recommend defining an exception hierarchy. The caller should be able to differentiate a logical error from a bug. Consider defining a base class for Kloppy errors, say `KloppyError` then additional classes derived from KloppyError to cover specific errors.

The Exception when pandas isn't present should probably remain an ImportError, but you can always add a more specfic message.

In general, a very positive update!

PR #119, opened 2021-11-29
Some bug fixes and improvements for Opta serializer

ThomasSeidl:
Hi,

1. fixed a bug where for incomplete passes end coordinates got ignored (within serializer and when converting events to_pandas)
2. added a result type "OWN_GOAL" for own goals - before these were just counted as normal goals for a team. I am wondering whether it would make sense to set BALL_OWNING_TEAM for this shot event to the opponent.
3. for all shot events the event timestamp will correspond to the moment when the ball leaves the foot. However for goal events opta's event timestamp is the moment when the ball crosses the goal line and the timestamp when the ball leaves the foot is stored in a qualifier. I adjusted the code s.t. it'll take always the moment when the ball leaves the foot consistently. This might not be a big deal when working with event data only, but a big improvement when working with merged tracking data.
4. added a parser for card events with event_type = "card" and column card_type = ["RED", "FIRST_YELLOW", "SECOND_YELLOW"]
5. starting player information will now get parsed

Cheers
Thomas

### DIFF
--- a/kloppy/domain/models/event.py
+++ b/kloppy/domain/models/event.py
@@ -230,6 +230,7 @@ class SetPieceQualifier(EnumQualifier):
 
     value: SetPieceType
 
+
 @dataclass
 class CardQualifier(EnumQualifier):
     """
@@ -240,6 +241,7 @@ class CardQualifier(EnumQualifier):
     """
 
     value: CardType
+
 
 class PassType(Enum):
     """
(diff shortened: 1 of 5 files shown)


koenvo:
This look great @ThomasSeidl !

Please make sure you also run code formatting: https://github.com/PySport/kloppy/blob/master/CONTRIBUTING.md#how-to-contribute item 5. If you have questions let me know.

PR #120, opened 2021-12-08
Bug fix when working with Opta Pass Qualifiers / converting those to_pandas

ThomasSeidl:
Hi,

based on some feedback I made some changes:

- opta pass_type wasn't working for a pass with multiple qualifiers (those got overwritten). Fixed this bug.
- Converting parsed events to pd.DataFrame using to_pandas() was also only able to keep the last pass type qualifier (in case there are multiple). Changed this similar to what you would get from [pd.get_dummies ](https://pandas.pydata.org/docs/reference/api/pandas.get_dummies.html), i.e. one column for each pass_type containing True/False, thus allowing one pass event to be a CROSS and ASSIST at the same time.
- added tests for both cases.

Cheers
Thomas

### DIFF
--- a/kloppy/domain/models/event.py
+++ b/kloppy/domain/models/event.py
@@ -230,6 +230,7 @@ class SetPieceQualifier(EnumQualifier):
 
     value: SetPieceType
 
+
 @dataclass
 class CardQualifier(EnumQualifier):
     """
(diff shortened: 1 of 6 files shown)


koenvo:
hi @ThomasSeidl, do you need some help here? If you like I can also have a look at the merge conflicts due to my latest refactoring.

ThomasSeidl:
Closing it, to build on top of a more up to date kloppy version in the future.

PR #126, opened 2022-01-24
Fix extraction of foul committed events

JanVanHaaren:
This pull request fixes an issue in the StatsBomb deserializer that prevents _foul committed_ events to be extracted from the StatsBomb event feed.

### Context
The StatsBomb deserializer attempts to convert each StatsBomb event into at most one corresponding `kloppy` event. The deserializer loops over the StatsBomb events (`raw_events`) and decides for each event (`raw_event`) whether it corresponds to one of the `kloppy` events. However, some StatsBomb events decompose into multiple `kloppy` events. For example, the _card_ event is a subtype of the _bad behaviour_ and _foul committed_ events in the StatsBomb event feed. Currently, the _card_ event is correctly extracted from the event feed, whereas the _foul committed_ event is ignored.

### Changes
1. The deserializer keeps a list of potentially new `kloppy` events while processing a `raw_event` from StatsBomb. This `new_events` list replaces the `event` variable that was used to keep a reference to the potentially new `kloppy` event. At the end of each iteration, the events that meet the inclusion criteria are added to the `kloppy` event feed.
2. The `event_type == SB_EVENT_TYPE_FOUL_COMMITTED` branch of the processing loop not only creates a _card_ event when a yellow or red card was given but also creates a _foul committed_ event.
3. The  `event_type == SB_EVENT_TYPE_BAD_BEHAVIOUR` and  `event_type == SB_EVENT_TYPE_FOUL_COMMITTED` branches of the processing loop have been refactored such that other properties could be extracted more conveniently.
4. The dict constructors have been replaced with dict literals, which are slightly faster and more flexible.

### DIFF
--- a/kloppy/infra/serializers/event/statsbomb/deserializer.py
+++ b/kloppy/infra/serializers/event/statsbomb/deserializer.py
@@ -179,12 +179,12 @@ def _parse_pass(pass_dict: Dict, team: Team, fidelity_version: int) -> Dict:
 
     qualifiers = _get_event_qualifiers(pass_dict)
 
-    return dict(
-        result=result,
-        receiver_coordinates=receiver_coordinates,
-        receiver_player=receiver_player,
-        qualifiers=qualifiers,
-    )
+    return {
+        "result": result,
+        "receiver_coordinates": receiver_coordinates,
+        "receiver_player": receiver_player,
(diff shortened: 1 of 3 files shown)


koenvo:
Thank you for this great work!

PR #133, opened 2022-03-17
Fix metadata update after dataset transformation

probberechts:
The metadata was not correctly updated after a dataset transformation.

- Pitch dimensions were set to None when transforming the orientation
- The coordinate system was not updated


koenvo:
Thanks for this PR! It would be great if you can add or adjust a test that breaks with previous implementation and is successful with your fix

probberechts:
I did that. See my changes to `kloppy/tests/test_helpers.py` where I extended the existing tests to make sure the added metadata is correct.

Or do you want me to split up the PR in two commits, one containing the failing tests and one containing the fix?

Issue #155, opened 2022-09-14
End coordinates missing for offside passes (Opta)


MKlaasman:
Not sure whether it is a design choice to not include end coordinates for offside passes. 

However, I suppose there are use-cases for wanting to investigate the direction & end location of an offside pass.
IMO the positives outweigh the negatives in terms of adding the end_coordinates in this use-case.

Background info:
Within the Opta data, offside passes have a different type_id (type_id == 2) compared to normal passes (type_id == 1).
Offside is thus not a subtype/qualifier of the main event pass. For this reason, they are handled differently by Kloppy:
- _parse_pass()
- _parse_offside_pass()

To do:
- add receiver_coordinates to _parse_offside_pass()

Not sure whether the same "issue" is present for other providers.

koenvo:
Thanks for solving this issue!

PR #156, opened 2022-09-14
Added end coordinates for offside passes within Opta deserializer

MKlaasman:
This Pull Request Includes:
- Opta deserializer now adds end coordinates to offside passes
- A test to check for end_coordinate_x value in Offside pass. (opta | type_id==2)

Fixes
Issue: #155


koenvo:
Thanks! Looking good.

Issue #158, opened 2022-10-04
Metrica EPTS loader doen't read the frameCount name from DataFormataSpecification


jcnunezmoreno:
When metadata.xml is loaded, the field name from frame count is hardcoded instead of read from the file.

koenvo:
Fixed in https://github.com/PySport/kloppy/pull/159

Issue #157, opened 2022-10-04
Metrica EPTS loader fails when sensor is defined but player channel doen't use this information


jcnunezmoreno:
EPTS loader from metrica provider is failing when a sensor is defined in the metadata.xml but player channel doen't contains information for this sensor. This occurs when any sensor want to be defined only for ball data.

koenvo:
Fixed in https://github.com/PySport/kloppy/pull/159

PR #159, opened 2022-10-04
Fixing metrica epts loader

jcnunezmoreno:
- Frame count name is load from xml file instead of use hardcoded value (#158)
- Including verification before indexing the sensor value to avoid crash  (#157)

### DIFF
--- a/kloppy/infra/serializers/tracking/metrica_epts/deserializer.py
+++ b/kloppy/infra/serializers/tracking/metrica_epts/deserializer.py
@@ -54,11 +54,11 @@ class MetricaEPTSTrackingDataDeserializer(
 
                 other_data = {}
                 for sensor in other_sensors:
+                    player_sensor_field_str = f"player_{player.player_id}_{sensor.channels[0].channel_id}"
+                    player_sensor_val = None if player_sensor_field_str not in row else row[player_sensor_field_str]
                     other_data.update(
                         {
-                            sensor.sensor_id: row[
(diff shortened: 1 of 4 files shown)


kloppy/infra/serializers/tracking/metrica_epts/deserializer.py
```diff
@@ -54,11 +54,11 @@ def _frame_from_row(
 
                 other_data = {}
                 for sensor in other_sensors:
+                    player_sensor_field_str = f"player_{player.player_id}_{sensor.channels[0].channel_id}"
+                    player_sensor_val = None if player_sensor_field_str not in row else row[player_sensor_field_str]
```
koenvo, line 58: `player_sensor_val = row.get(player_sensor_field_str)`

koenvo (commented):
Thanks for this PR!

Could you update a Metrica EPTS test so it will test this behavior? And one minor code improvement comment.

kloppy/infra/serializers/tracking/metrica_epts/deserializer.py
jcnunezmoreno: Done!

koenvo:
Sorry for the delay. Thanks for the changes! Looking good to me. Can you also run code formatting?

jcnunezmoreno:
> Sorry for the delay. Thanks for the changes! Looking good to me. Can you also run code formatting?

Done! Thanks!!

PR #160, opened 2022-10-12
Feature/event factory

koenvo:
Fixes https://github.com/PySport/kloppy/issues/53

The fix makes it possible to extend the `Event` classes to contain vendor specific information like `expected_goals`.


JanVanHaaren (approved):
This pull request looks good to me! The event factory is a really nice addition. Great stuff.

I did notice one minor inconsistency in the use of the StatsBomb company name in the names for classes: `StatsBombDeserializer`, `StatsbombInputs`, `StatsbombPassEvent`. I do not have a preference, but I would be consistent to avoid confusion.

PR #164, opened 2022-11-23
Freeze frame support

koenvo:
This PR adds support for freeze frames. The implementation is vendor agnostic. This means the freeze frame information can be loaded from any tracking data source. In this PR data is loaded from StatsBomb event data and 360 data. When you want to use a different source you need to think about how to sync the data frame accurate.

Several changes are made:
1. A `freeze_frame` attribute of type `Frame` is added to `Event`
2. The StatsBomb deserializer now loads freeze frames from event data
3. When a 360 feeds is passed to the StatsBomb deserializer the data is attached to the events
4. The Transformer is changed to also transform freeze frames within event data
5. The 'Event data' getting started notebook is extended with a freeze frame example


kloppy/infra/serializers/event/statsbomb/deserializer.py
```diff
@@ -319,10 +315,64 @@ def _parse_shot(shot_dict: Dict) -> Dict:
...
+                freeze_frame_player["location"], fidelity_version
+            )
+        )
+
+    if event.player not in players_data:
+        # Shot data from event stream might not include coordinates for event player
+        players_data[event.player] = PlayerData(coordinates=event.coordinates)
+
+    return Frame(
+        frame_id=None,  # TODO: come up with a clever value here
```
JanVanHaaren, line 365: How about using the `event_uuid` value, potentially with a prefix or suffix?

JanVanHaaren (approved):
This pull request looks good to me. Nice addition!

kloppy/infra/serializers/event/statsbomb/deserializer.py
koenvo: Was thinking about it. Now it's of type `int`. Maybe I should calculate it from the timestamp with a assumed fps of 25?

koenvo: Maybe `(event.period.end_timestamp + event.timestamp) * FPS` with `FPS = 25`

Issue #167, opened 2023-02-07
Metrica EPTS metadata deserializer does not set orientation


probberechts:
The orientation and attacking direction is not set (correctly) in the metadata deserializer of the Metrica EPTS format:

1. The deserializer determines the orientation based on the attacking direction attribute of the match periods ([code](https://github.com/PySport/kloppy/blob/d523b5d44d27cf639f99c3c340d3da9e3611b352/kloppy/infra/serializers/tracking/metrica_epts/metadata.py#L258)).  However, the result is not added to the final ` EPTSMetadata` object ([code](https://github.com/PySport/kloppy/blob/d523b5d44d27cf639f99c3c340d3da9e3611b352/kloppy/infra/serializers/tracking/metrica_epts/metadata.py#L288)).
2. The attacking direction is not extracted from the metadata files ([code](https://github.com/PySport/kloppy/blob/d523b5d44d27cf639f99c3c340d3da9e3611b352/kloppy/infra/serializers/tracking/metrica_epts/metadata.py#L61)). It is included in the raw data though. E.g., for [sample match 3](https://github.com/metrica-sports/sample-data/blob/master/data/Sample_Game_3/Sample_Game_3_metadata.xml):
```xml
<Teams>
  <Team id="FIFATMA">
    <Name>Team A</Name>
    <ProviderTeamsParameters>
      <ProviderParameter>
        <Name>attack_direction_first_half</Name>
        <Value>right_to_left</Value>
      </ProviderParameter>
    </ProviderTeamsParameters>
  </Team>
  <Team id="FIFATMB">
    <Name>Team B</Name>
    <ProviderTeamsParameters>
      <ProviderParameter>
        <Name>attack_direction_first_half</Name>
        <Value>left_to_right</Value>
      </ProviderParameter>
    </ProviderTeamsParameters>
  </Team>
</Teams>
```
3. The orientation is set to `FIXED_...` ([code](https://github.com/PySport/kloppy/blob/d523b5d44d27cf639f99c3c340d3da9e3611b352/kloppy/infra/serializers/tracking/metrica_epts/metadata.py#L260)) but at least for the Metrica sample data (match 3) it seems to alternate between periods. 
```py
>> from kloppy.metrica import load_open_data

>> dataset = load_open_data("3")
>> df = dataset.to_pandas()
>> # avg position of home goalie
>> df.groupby("period_id")["P3578_x"].mean()
period_id                                                                                                                              
1    0.917014                                                                                                                          
[...]

PR #168, opened 2023-02-07
Set orientation for Metrica EPTS (#167)

probberechts:
**DISCLAIMER**: this is based on sample match 3 in the Metrica open data. I do not know whether this solution generalizes.

### DIFF
--- a/kloppy/infra/serializers/tracking/metrica_epts/metadata.py
+++ b/kloppy/infra/serializers/tracking/metrica_epts/metadata.py
@@ -39,10 +39,16 @@ def _load_provider_parameters(parent_elm, value_mapper=None) -> Dict:
     }
 
 
-def _load_periods(global_config_elm, frame_rate: int) -> List[Period]:
+def _load_periods(metadata_elm, team_map: dict, frame_rate: int) -> List[Period]:
+    global_config_elm = metadata_elm.find("GlobalConfig")
     provider_params = _load_provider_parameters(
         global_config_elm.find("ProviderGlobalParameters")
     )
+
+    provider_teams_params = {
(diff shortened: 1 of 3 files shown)


koenvo:
Thanks for this PR!

@jcnunezmoreno are you able to verify this fix generalizes?

jcnunezmoreno:
@probberechts Thanks for the PR! The problem is FIFA EPTS only defines how to structure the data but not what you can find there, this is the reason why this EPTS is related to Metrica, because is the way we export the data.

@koenvo Related to you question, we always use the keys/values for the orientation as in the sample 3, so it will generalizes for the rest of the files. But I tested on my own sample game and the orientation was not set, it contains the value "not-set", maybe a test can be added based on the sample game 3. If it works I can share with you other EPTS file (the one that is not working for me) and try to add to the test.

probberechts:
> @probberechts Thanks for the PR! The problem is FIFA EPTS only defines how to structure the data but not what you can find there, this is the reason why this EPTS is related to Metrica, because is the way we export the data.

It is clear that this is specifically a parser for the **Metrica** EPTS format. So, then I think this PR is fine.

> But I tested on my own sample game and the orientation was not set, it contains the value "not-set", maybe a test can be added based on the sample game 3. 

There are tests included on a metadata file that has the same structure as sample game 3. I also tested it on sample game 3 and it returns the correct orientation for me (see example below). Hence, I am curious what's the difference with your sample game. 

```py
>> from kloppy import metrica
>> dataset = metrica.load_open_data(match_id="3", limit=10)
>> dataset.metadata.orientation
away-team
```

jcnunezmoreno:
@probberechts Yeah, sorry, you are right! I was checking different field. I checked it on some games and it works good! Thanks!

Issue #169, opened 2023-02-14
Support Wyscout event data v3


DriesDeprest:
Kloppy currently only supports loading Wyscout v2 event data, while Wyscout v3 event data is the new standard. It would be great to add support for loading Wyscout v3 data.

![image](https://user-images.githubusercontent.com/46134788/218786128-770f900b-8fc1-4b42-9c5a-e72537d5dbf8.png)

JanVanHaaren:
Yes, I agree that adding support for the V3 specification of the Wyscout event data would be great.

koenvo:
Fixed in https://github.com/PySport/kloppy/pull/170. Thanks a lot!

PR #170, opened 2023-02-14
Add support for Wyscout V3

DriesDeprest:
Created a PR related to issue described in https://github.com/PySport/kloppy/issues/169


JanVanHaaren:
Thank you for the pull request, @DriesDeprest.

I haven't had a chance to study the changes in detail yet. However, I would suggest keeping the support for the V2 specification of the Wyscout event data for now. Although the V2 specification may be outdated, I think the corresponding deserializer still plays a valuable role in kloppy. Moreover, the deserializer is currently used in existing projects, and [one of the largest publicly available event datasets](https://figshare.com/collections/Soccer_match_event_dataset/4415000/5) still employs the V2 specification. If we were to remove the support for V2, users would no longer be able to process that dataset with kloppy.

DriesDeprest:
Hi Jan,

Thanks for the feedback, I agree. On a high level, what do you think would be a good implementation to both support V2 and V3 Wyscout events data?

I think the most elegant would be to:
- have an infra/serializers/event/wyscout/deserializer_v2.py for V2, which has a WyscoutDeserializerV2 class, which contains the old code supporting v2
- have an infra/serializers/event/wyscout/deserializer_v3.py for V3, which has a WyscoutDeserializerV3 class, which contains my new code
- Add a version argument (default value = v3) to wyscout.load method and conditional behiavour based on the argument to use the V2 or V3 WyscoutDeserializer?

Another approach of trying to keep a single infra/serializers/event/wyscout/deserializer.py and update all the functions inside it to have conditional behaviour based on V2 and V3 would result in less readable and elegant code imo. 

Also, over time, if we remove support for V2 and add support for a future version e.g. v4 it will be an easy update to the library of just removing the deserializer_v2.py file, adding a deserializer_v4.py file and updating the version argument handling in the wyscout.load method.

Wdyt?

DriesDeprest:
Hi @JanVanHaaren, did you have a chance to consider my approach suggested in my comment above? Just wanted to check whether you agree, before I start implementing it.

JanVanHaaren:
Thank you, @DriesDeprest. I've discussed your proposal with @koenvo, and we generally like it. However, we want to ensure backward compatibility with existing codebases that rely on kloppy. If version 3 became the default version, many existing projects would break. Could we explore the possibility of automatically detecting the version if no version is explicitly provided to the deserializer?

kloppy/_providers/wyscout.py
```diff
@@ -23,16 +26,30 @@ def load(
...
     """
-    deserializer = WyscoutDeserializer(
-        event_types=event_types,
-        coordinate_system=coordinates,
-        event_factory=event_factory or get_config("event_factory"),
-    )
-    with open_as_file(event_data) as event_data_fp:
-        return deserializer.deserialize(
-            inputs=WyscoutInputs(event_data=event_data_fp),
+    if not data_model_version:
```
JanVanHaaren, line 31: I'm wondering whether we can somewhat harden the implementation of this function. I see two potential issues with the current implementation.

1. The function won't do anything when a user provides a string that is neither "V3" nor "V2" for the `data_model_version` parameter. This issue could be avoided by calling the `identify_data_model_version` function if a user provides an argument for the `data_model_version` parameter that is neither "V3" nor "V2".
2. The `if` and `elif` branches contain a lot of duplicated code. The only difference between the two branches is the name of the deserializer class. This issue could be avoided by storing the class in the `deserializer` variable. In addition, the `identify_data_model_version` function could also return the appropriate class.

kloppy/_providers/wyscout.py
```diff
@@ -47,3 +64,19 @@ def load_open_data(
         coordinates=coordinates,
         event_factory=event_factory,
     )
+
+
+def identify_data_model_version(event_data):
```
JanVanHaaren, line 69: This function does not have type hints.

kloppy/_providers/wyscout.py
```diff
@@ -47,3 +64,19 @@ def load_open_data(
...
+    with open(event_data) as json_file:
+        events = json.load(json_file)
+    events = events["events"]
+    if "eventName" in events[0]:
```
JanVanHaaren, line 73: I would store `events[0]` in a variable named `first_event` or something along those lines to improve the readability of the code.

kloppy/_providers/wyscout.py
```diff
@@ -47,3 +64,19 @@ def load_open_data(
...
+        data_model_version = "V2"
+        return data_model_version
+    elif ("type" in events[0]) and ("primary" in events[0]["type"]):
```
JanVanHaaren, line 76: How about `elif "primary" in first_event.get("type", {})`?

kloppy/_providers/wyscout.py
```diff
@@ -14,6 +16,7 @@ def load(
     event_types: Optional[List[str]] = None,
     coordinates: Optional[str] = None,
     event_factory: Optional[EventFactory] = None,
+    data_model_version: Optional[str] = None,
```
JanVanHaaren, line 19: Did you include the term `model` in the variable name on purpose? I feel that it doesn't add much and can perhaps even be confusing in the context of football analytics, where the term _model_ is often used for different purposes. How about `data_version` or `data_specification`?

kloppy/infra/serializers/event/wyscout/deserializer_v3.py
```diff
@@ -0,0 +1,394 @@
...
+            result = TakeOnResult.INCOMPLETE
+
+    return {"result": result, "qualifiers": qualifiers}
+
+
+def _players_to_dict(players: List[Player]):
+    return {player.player_id: player for player in players}
+
+
+class WyscoutInputs(NamedTuple):
```
JanVanHaaren, line 257: Do we need to duplicate this class?

kloppy/_providers/wyscout.py
```diff
@@ -47,3 +64,19 @@ def load_open_data(
         coordinates=coordinates,
         event_factory=event_factory,
     )
+
+
+def identify_data_model_version(event_data):
+    with open(event_data) as json_file:
+        events = json.load(json_file)
+    events = events["events"]
```
JanVanHaaren, line 72: I'd avoid re-using the `events` variable name.

JanVanHaaren (commented):
This pull request generally looks good to me. I've added a few suggestions, mostly to avoid code duplication and improve readability. Thank you for your contribution!

DriesDeprest:
Thanks for the feedback @JanVanHaaren, with my latest commit, your comments should now be resolved. Either let me know if this is not the case, or feel free to make extra commits to this PR, to make it merge-ready.

JanVanHaaren:
Thank you, @DriesDeprest! I don't have write permissions to the branch, but I'd still make the following changes. I haven't extensively tested the implementation so please let me know if you see any potential issues that I may have overlooked.

```python
def load(
    event_data: FileLike,
    event_types: Optional[List[str]] = None,
    coordinates: Optional[str] = None,
    event_factory: Optional[EventFactory] = None,
    data_version: Optional[str] = None,
) -> EventDataset:
    if data_version == "V2":
        deserializer_class = WyscoutDeserializerV2
    elif data_version == "V3":
        deserializer_class = WyscoutDeserializerV3
    else:
        deserializer_class = identify_deserializer(event_data)

    deserializer = deserializer_class(
        event_types=event_types,
        coordinate_system=coordinates,
        event_factory=event_factory or get_config("event_factory"),
    )

    with open_as_file(event_data) as event_data_fp:
            inputs=WyscoutInputs(event_data=event_data_fp),
        )
```

```python
def identify_deserializer(event_data: FileLike) -> str:
    with open(event_data) as json_file:
        events_with_meta = json.load(json_file)

    events = events_with_meta["events"]
    first_event = events[0]

    deserializer = None
    if "eventName" in first_event:
        deserializer = WyscoutDeserializerV2
    elif "primary" in first_event.get("type", {}):
        deserializer = WyscoutDeserializerV3

    if deserializer is None:
        raise ValueError(
            "Wyscout data version could not be recognized, please specify"
        )

    return deserializer
```

koenvo:
One small note: make sure to use `open_as_file` instead of plain `open`. The input can be some different types which `open_as_file` takes care of.

DriesDeprest:
@JanVanHaaren thanks for the additional input, I've integrated them in my last commit. Let me know if additional changes are required.

JanVanHaaren:
Thanks. The pull request looks good to me now!

DriesDeprest:
Great @JanVanHaaren, thanks for all the feedback on this PR! What is the next step for getting the PR merged? Approval of @koenvo ?

Also, I'm quite eager to start using this in one of my projects. So I'm also wondering when we can have a new release with this PR included, do you have any idea about that?

Let me know if I can help in any way, thanks in advance! :-)

koenvo:
> Great @JanVanHaaren, thanks for all the feedback on this PR! What is the next step for getting the PR merged? Approval of @koenvo ?
> 
> Also, I'm quite eager to start using this in one of my projects. So I'm also wondering when we can have a new release with this PR included, do you have any idea about that?
> 
> Let me know if I can help in any way, thanks in advance! :-)

Thanks a lot both! Let me do a final check, and then merge and release.

DriesDeprest:
Sorry for my impatience, but did you already find the time to review the PR @koenvo? Thanks in advance!

koenvo:
> Sorry for my impatience, but did you already find the time to review the PR @koenvo? Thanks in advance!

Can you run code formatting and push again please?

koenvo:
Thanks a lot Dries!

PR #172, opened 2023-03-03
Add polars support

koenvo:
Main additional of the PR is the ability to export a `Dataset` to a polars `DataFrame`. The refactor required to make this possible, makes it also possible to add different DataFrame engines like veax.

Also the `to_pandas` methods are deprecated in favour of `to_df`.


JanVanHaaren (approved):
Looks good!

PR #174, opened 2023-03-03
Fix for issue 171

### LINKED ISSUE #171: Incorrect home/away mapping in Statsbomb metadata
When working with Statsbomb data I've noticed that the home and away teams are not always parsed correctly. This is probably because they are loaded from lineup data.

https://github.com/PySport/kloppy/blob/abd9af208bbf0505731bc9ef21a9778c5a8317e9/kloppy/infra/serializers/event/statsbomb/deserializer.py#L508

The match files provided by Statsbomb with detailed match metadata should probably be used instead. Here are some examples of discrepancies (in my little experience this happens about every second case):

image

```
import requests
from kloppy import statsbomb
from kloppy.domain.models import Ground

response = requests.get("https://raw.githubusercontent.com/statsbomb/open-data/master/data/matches/11/1.json")
matches = response.json()

count = 0
for match in matches:
    dataset = statsbomb.load_open_data(str(match['match_id']))
    parsed_home_team = [team.team_id for team in dataset.metadata.teams if team.ground == Ground.HOME][0]
    if str(match['home_team']['home_team_id']) != str(parsed_home_team):
        print("Match ID: ", match['match_id'])    
        print("home team from matches file: ", match['home_team']['home_team_id'])
        print("home team stated in metadata: ", parsed_home_team)
        print('\n')
        count += 1
```

koenvo:
Closes #171

### DIFF
--- a/kloppy/infra/serializers/event/statsbomb/deserializer.py
+++ b/kloppy/infra/serializers/event/statsbomb/deserializer.py
@@ -505,7 +505,14 @@ class StatsBombDeserializer(EventDataDeserializer[StatsBombInputs]):
 
         with performance_logging("load data", logger=logger):
             raw_events = json.load(inputs.event_data)
-            home_lineup, away_lineup = json.load(inputs.lineup_data)
+            lineups = json.load(inputs.lineup_data)
+
+            # First event (Starting-XI) determines home team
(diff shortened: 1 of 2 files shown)


JanVanHaaren (approved):
Looks good!

PR #179, opened 2023-03-25
Fix S3Adapter.read_to_stream

probberechts:
The S3Adapter used shutil.copyfile to copy the contents of an open file handle to a buffer. However, the arguments to copyfile() should be file names. The correct method is shutil.copyfileobj() which operates on open file handles.


koenvo:
Thanks for the fix!

PR #180, opened 2023-04-11
Add pandas pyarrow support + better performance on to_df

koenvo:
This PR has two main changes:
1. The `Dataset.to_df` formats the data as `Dict[str, List[Any]]` instead of `List[Dict[str, Any]]` before passing it to the DataFrame constructors. This makes it a lot more performant to create a DataFrame, as the schema is already known (?)
2. Added support for `pandas[pyarrow]` engine.

### DIFF
--- a/kloppy/domain/models/common.py
+++ b/kloppy/domain/models/common.py
@@ -1,5 +1,6 @@
 import sys
 from abc import ABC, abstractmethod
+from collections import defaultdict
 from dataclasses import dataclass, field, replace
 from enum import Enum, Flag
 from typing import (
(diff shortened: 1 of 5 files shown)


kloppy/domain/models/common.py
```diff
@@ -940,28 +941,92 @@ def to_records(
         else:
             return iterator
 
+    def to_dict(
```
JanVanHaaren, line 944: Detail: I'm wondering whether the `to_dict` method name could be potentially confusing. Intuitively, without looking at the actual implementation, I didn't expect this method to return a `dict` where each value corresponds to a list of values by default. For instance, the `pandas.DataFrame` class also offers a `to_dict` method which shows the same behavior when passing the `list` argument to the `orient` parameter.

kloppy/domain/models/common.py
```diff
@@ -940,28 +941,92 @@ def to_records(
...
     ):
         from kloppy.config import get_config
 
         if not engine:
             engine = get_config("dataframe.engine")
 
-        if engine == "pandas":
-            from pandas import DataFrame
+        if engine == "pandas[pyarrow]":
+            try:
```
JanVanHaaren, line 983: Do we need this `try-except` construct when the `engine` is set to `pandas[pyarrow]`? Wouldn't it be enough to check whether `ArrowDtype` can be imported from `pandas`?

JanVanHaaren (approved):
This pull request looks good to me! I love the PyArrow support.

JanVanHaaren (approved):

PR #182, opened 2023-04-20
Rishicarter/issue60 remove deleted events

rishicarter:
This PR is for removing the 'deleted events' from the Opta f24 dataset. 

As per the documentation of Opta's F24 match feed data, deleted events are associated with type_id='43' and a qualifier_id='144'. As per my understanding, once an event is deleted, Opta does remove the associated event's data but the value of the event_id attribute remains in the feed with type id='43'. An example of a deleted event:

image

Source: [Link](https://github.com/jokecamp/FootballData/blob/master/random_docs/Opta-f24_appendices.docx)

I also checked the dict event_type_names in the deserializer.py file and found id - '43' to be associated with 'deleted events'. I believe a possible fix can be to skip the deleted events completely. If this sounds reasonable, I would be happy to work on implementing this fix.

Code change:
Skipping events that start with type id '43' for Opta.


kloppy/tests/issues/issue_60/test_issue_60.py
```diff
@@ -0,0 +1,45 @@
...
+        },
+    )
+
+    metadata = dataset.metadata
+    home_team, away_team = metadata.teams
+    return events, home_team.team_id, away_team.team_id
+
+
+class TestIssue60:
+    def test_parse_opta(self):
```
koenvo, line 40: Please make sure in this test that the dataset does not contain the deleted events, but does include the other ones.

rishicarter: Hi Koenvo, I updated Optaf24 to contain deleted events. This was done to test if my changes were working. I am not quite sure why it needs to be removed. Sorry to be a bother.

koenvo: First: thanks for this PR and sorry for the delay. 

Second: I understand I wasn't very clear with my comment. Please make sure in your test that you test the dataset does not contain any events that should be filtered out because they are deleted.

You can find an example here: https://github.com/PySport/kloppy/blob/master/kloppy/tests/test_event.py#L55-L66

PR #183, opened 2023-04-28
Allow chaining of operators on a Dataset (filter + map). 

koenvo:
This PR includes:
1. A `map` method on the `Dataset` to allow chaining of operators: filter/map
2. Better representation of `Dataset` to not break IDE/Jupyter when you do `print(dataset)` or display it in any other way
3. Some fixes for formations. This is based on loading the entire Statsbomb open data repository.

This should have been multiple PRs but it kind of slipped in.


JanVanHaaren (approved):
Looks good to me!

Issue #184, opened 2023-05-02
Support StatsPerform Tracking data


MKlaasman:
**Description**
There is no implementation for loading StatsPerform tracking data. To allow Kloppy to load this data, there should be added an extra deserializer for StatsPerform. 

**Proposal**

- Create _serializers/tracking/statsperform.py_ : that handles the loading of StatsPerform tracking data with input files being: metadata.xml, players.csv & tracking.txt
- Add _test_statsperform.py_
- Add StatsPerform coordinate system to _common.py_  (x: 0,100 & y: 0,100)

JanVanHaaren:
Hi Milan. I like the idea of adding a deserializer for Stats Perform!

1. I believe that Stats Perform are currently using slightly different setups in different leagues. Do I understand correctly that your proposal refers to SportVU 2.0? If so, I would be in favor of being more explicit about it.
2. I am a bit confused regarding the three required input files. Based on the pull request, the `metadata.xml` file appears to contain the MA25 feed from the Stats Perform API, whereas the `players.csv` and `tracking.txt` files appear to be in the Stats format. Is there a chance that we could rely entirely on responses from the Stats Perform API? For example, take the metadata from the MA1 feed and the tracking data from the MA25 feed?

denizcanoruc:
Hi Milan @MKlaasman and Jan @JanVanHaaren ,

I have already implemented both tracking and event data serializer for StatsPerform for a project of mine. But it has some minor bugs right now I need to fix. If it fits you, I can share it with you after fixing the bugs. It can take about 2 days.

MKlaasman:
@denizcanoruc Requirements here are that it fits in the Kloppy structure. 

If you think it can be beneficial for the implementation, you can always reach-out to me via LinkedIn or Twitter.

JanVanHaaren:
While reviewing pull request #191, I was wondering the following: What would be the best way to deal with data providers that provide their data in multiple formats? For example, the Stats Perform API supports both XML and JSON, but the current implementation of the deserializer only supports XML. Should we make this more explicit by, for example, including XML in the name of the deserializer class? Any thoughts, @koenvo?

koenvo:
### Be explicit

Being explicit and have 'xml' or something in the class name (and corresponding function in `kloppy/<vendor>.py`) makes sense. People will often know in what format they have the files. And especially when `kloppy` is used within a bigger python program, the program should know the format upfront.

### or Auto-detect

[...]

PR #185, opened 2023-05-02
Feature/add support for statsperform tracking data

MKlaasman:
Created a PR related to issue described in #184


koenvo:
Fixed in https://github.com/PySport/kloppy/pull/191

PR #189, opened 2023-05-11
Fix datatype of SkillCorner metadata.periods

probberechts:
The SkillCorner deserializer stored the `metadata.periods` attribute as a dictionary. According to the spec, it should be stored as a list of periods.


koenvo:
Thanks a lot Pieter!

PR #191, opened 2023-05-15
enable statsperform tracking data

MKlaasman:
Second implementation of issue #184 
Now using the following input files:
- MA1: metadata  (including line-ups)
- MA25: raw_data

@JanVanHaaren can you take a look, whether this implementation is what you expected?


kloppy/infra/serializers/tracking/statsperform.py
```diff
@@ -0,0 +1,271 @@
...
+    ):
+        super().__init__(limit, sample_rate, coordinate_system)
+        self.only_alive = only_alive
+
+    @property
+    def provider(self) -> Provider:
+        return Provider.STATSPERFORM
+
+    @classmethod
+    def __get_periods(cls, tracking):
```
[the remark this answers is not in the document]
JanVanHaaren:
```python
  @classmethod
  def __get_periods(cls, tracking):
      """Gets the Periods contained in the tracking data."""
      lines = tracking.decode("ascii").splitlines()
      
      period_data = {}
      for line in lines:
          time_info = line.split(";")[1].split(",")
          period_id = int(time_info[1])
          frame_id = int(time_info[0])
          if period_id not in period_data:
              period_data[period_id] = set()
          period_data[period_id].add(frame_id)

      periods = [
          Period(
              id=period_id,
              start_timestamp=min(frame_ids),
              end_timestamp=max(frame_ids),
          )
          for period_id, frame_ids in period_data.items()
      ]

      return periods
```

kloppy/infra/serializers/tracking/statsperform.py
```diff
@@ -0,0 +1,271 @@
...
+        return periods
+
+    @classmethod
+    def __get_frame_rate(cls, tracking):
+        """gets the frame rate of the tracking data"""
+        lines = tracking.decode("ascii").splitlines()
+        timestamps = []
+        num_frames = 5
+
+        for i in [0, num_frames]:
```
JanVanHaaren, line 88: I would use a `tuple` instead of `list` in this situation. I was wondering whether we really need a loop, but I figured that it probably doesn't really matter in terms of readability and performance.

I am wondering though whether the current implementation is sufficiently robust. Is it fair to assume that the data is always sorted in this specific way and that no frames are missing?

kloppy/infra/serializers/tracking/statsperform.py
```diff
@@ -0,0 +1,271 @@
...
+        frame_id = int(frame_info[0])
+        frame_timestamp = int(frame_info[1].split(",")[0]) / 1000
+        match_status = int(frame_info[1].split(",")[2])
+
+        ball_state = BallState.ALIVE if match_status == 0 else BallState.DEAD
+        ball_owning_team = None
+
+        if len(components) > 2:
+            ball_x, ball_y, ball_z = map(
+                float, components[2].split(";")[0].split(",")
```
JanVanHaaren, line 111: I would store the result of the `components[2].split(";")[0].split(",")` statement in an auxiliary variable to improve the readability of the code.

kloppy/infra/serializers/tracking/statsperform.py
```diff
@@ -0,0 +1,271 @@
...
+            (
+                team_side_id,
+                player_id,
+                jersey_no,
+                x,
+                y,
+            ) = player_data.split(",")
+
+            # Goalkeepers have id 3 and 4
+            if int(team_side_id) > 2:
```
JanVanHaaren, line 129: I would store the result of the `int(team_side_id)` cast in an auxiliary variable to improve the readability of the code.

kloppy/infra/serializers/tracking/statsperform.py
```diff
@@ -0,0 +1,271 @@
...
+        players_data = {}
+        player_info = components[1]
+        for player_data in player_info.split(";")[:-1]:
+            (
+                team_side_id,
+                player_id,
+                jersey_no,
+                x,
+                y,
+            ) = player_data.split(",")
```
JanVanHaaren, line 126: I would do all the required type casts immediately after the `player_data.split(",")` call to avoid cluttering the "business logic" with type casts.

kloppy/infra/serializers/tracking/statsperform.py
```diff
@@ -0,0 +1,271 @@
...
+            pitch_size_length = 100
+            pitch_size_width = 100
+
+            periods = self.__get_periods(raw_data)
+            # Get Player Info:
+            for line_up in objectify.fromstring(
+                metadata
+            ).liveData.iterchildren(tag="lineUp"):
+                count = 0
+                for team in teams:
```
JanVanHaaren, line 194: I would use `enumerate` instead of manually keeping track of the `count` variable.

However, I think an even better solution would be to turn the `teams` `list` into a `dict` where the key is the team ID and the value is the `Team` object.

```python
team_id = line_up.get("contestantId")
team = teams[team_id]

teams_list = list(teams.values())
metadata = Metadata(
    teams=teams_list,
    ...
)
```

kloppy/infra/serializers/tracking/statsperform.py
```diff
@@ -0,0 +1,271 @@
...
+                length=pitch_size_length, width=pitch_size_width
+            )
+
+            def _iter():
+                n = 0
+                sample = 1.0 / self.sample_rate
+
+                for line_ in raw_data.decode("ascii").splitlines():
+                    frame_id = int(line_.split(";")[1].split(",")[0])
+                    if self.only_alive and not line_.endswith("Alive;:"):
```
JanVanHaaren, line 224: What does the `not line_.endswith("Alive;:")` test do exactly?

kloppy/infra/serializers/tracking/statsperform.py
```diff
@@ -0,0 +1,271 @@
...
+                        continue
+                    period_id = int(line_.split(";")[1].split(",")[1])
+                    period_ = periods[period_id - 1]
```
JanVanHaaren, line 227: How about letting the `__get_periods` class method return a `dict` instead of a `list` such that the `Period` object could be retrieved by its period ID? I would avoid relying on list indexes whenever possible.

kloppy/infra/serializers/tracking/statsperform.py
```diff
@@ -0,0 +1,271 @@
...
+                frames.append(frame)
+
+                if not period.attacking_direction_set:
+                    period.set_attacking_direction(
+                        attacking_direction=attacking_direction_from_frame(
+                            frame
+                        )
+                    )
+
+                if self.limit and n + 1 >= self.limit:
```
JanVanHaaren, line 247: Small suggestion to improve the readability: Add `start=1` to the `enumerate` call and change the implementation to `and n >= self.limit`.

kloppy/tests/files/statsperform_MA1_metadata.xml
JanVanHaaren, line 1: Perhaps we should change `MA1` to `ma1` in the filename for consistency with the Opta F7 and F24 files?

kloppy/tests/files/statsperform_MA25_tracking.txt
JanVanHaaren, line 1: Perhaps we should change `MA25` to `ma25` in the filename for consistency with the Opta F7 and F24 files?

kloppy/_providers/statsperform.py
```diff
@@ -0,0 +1,32 @@
...
+from kloppy.infra.serializers.tracking.statsperform import (
+    StatsperformDeserializer,
+    StatsperformInputs,
+)
+from kloppy.io import FileLike, open_as_file
+
+
+def load(
+    meta_data: FileLike,  # Statsperform MA1 file - XML - single game, live data & lineups
+    raw_data: FileLike,  # Statsperform MA25 file
```
JanVanHaaren, line 13: I think we should try to make more explicit that the `raw_data` file is a text file rather than an XML file. The MA25 API endpoint returns a response in either XML or JSON format, which includes a download link to the text files that contain the actual tracking data.

JanVanHaaren (commented):
I think this pull request generally looks good already! I have added a few suggestions that could help to improve the readability of the code.

kloppy/infra/serializers/tracking/statsperform.py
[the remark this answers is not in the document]
MKlaasman: Thanks! Found myself struggling to do this elegantly

kloppy/infra/serializers/tracking/statsperform.py
MKlaasman: Currently, all deserializers are using a `list` for the periods variable. Making the logic to retrieve the first half period as follows: `periods[0]`.

If we would want to change the logic to allow for using a `dict`, I would propose to create a new `issue` to change this for all deserializers.

kloppy/infra/serializers/tracking/statsperform.py
MKlaasman: In the new implementation, I calculate the most occurring frame delta (https://github.com/PySport/kloppy/pull/191/commits/f80f3ef35b99e9b2458c13a266b576a94a8beab5). Let me know, whether you think this is robust enough. 

Par exemple:
frames: [0, 100, 200, 500]
deltas: [100, 100, 300]
Result in a found delta of 100 (n=2).

kloppy/tests/test_statsperform.py
```diff
@@ -0,0 +1,107 @@
...
+    DatasetType,
+)
+
+from kloppy import statsperform
+
+
+class TestStatsperformTracking:
+    @pytest.fixture
+    def meta_data(self) -> str:
+        base_dir = os.path.dirname(__file__)
```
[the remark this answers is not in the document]
MKlaasman: Added this in the newest implementation, also added the conftest.py. (https://github.com/PySport/kloppy/pull/191/commits/f80f3ef35b99e9b2458c13a266b576a94a8beab5#diff-91d9d654856f8e3e615cf0cd03d0853c6b944aa21d962d399d406f081ad32d2eL18-R25)

MKlaasman:
@JanVanHaaren The newest commit should cover all your remarks. Please let me know, if I should change something!

MKlaasman:
@JanVanHaaren Thanks for your feedback. The newest implementation should answer all of your remarks. 

Let me know, if I should improve something.

JanVanHaaren (approved):
Thanks for the changes! This pull request looks good to me now.

Issue #173, opened 2023-03-03
Use Pathlib instead of str path in tests


koenvo:
Most tests use string paths with f-strings to concat them. This should be replaced with Pathlib

PR #193, opened 2023-05-24
Refactor pathlib tests

plaguss:
This pull request tries to solve the issue [#173 ](https://github.com/PySport/kloppy/issues/173).
It includes changes in multiple files, but they are all related to the issue :).
Also, given `base_dir` is a fixture reused across a lot of tests, a `conftest.py` file was added (just for this variable, I don't know if it would be interesting to add other fixtures reused in here).
It also includes a change (an update of the black version) to `.pre-commit-config.yaml` that I noticed when running the pre-commit hook.
If there are other changes expected to close the issue, please let me know, I'd be happy to help.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
 -   repo: https://github.com/ambv/black
-    rev: 20.8b1
+    rev: 23.3.0
     hooks:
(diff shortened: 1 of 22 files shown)


koenvo:
Thanks a lot for this contribution!

Issue #138, opened 2022-03-18
Add a data model for keeper actions


probberechts:
Kloppy's event data model currently does not support keeper actions. It would be great if this could be added.

### How data providers do it

**StatsBomb** has a single "Goal Keeper" event where the type of the intervention is coded as "Collected", "Goal Conceded", "Keeper Sweeper", "Penalty Conceded", "Penalty Saved", "Punch", "Penalty Saved To Post", "Save", "Saved to Post", "Shot Faced", "Shot Saved", "Shot Saved Off T", "Shot Saved To Post" and "Smother".

**Opta** has seven keeper events: "Save", "Claim", "Punch", "Cross not claimed", "Smother", "Keeper pick-up", and "Penalty faced".

**Wyscout** has "Shot against (Save attempt)" and "Goalkeeper leaving line" events.

**SoccerAction** maps keeper events to four actions: "keeper_save", "keeper_claim", "keeper_punch" and "keeper_pick_up".

### Proposal

I suggest creating a single `GoalkeeperEvent` and use `GoalkeeperQualifiers` to distinguish between a `Save`, `Claim`, `Punch`, `PickUp` and `Smother`.

joemulberry:
Is this widely accepted as an agreed plan?

MKlaasman:
@koenvo & @JanVanHaaren do you see any caveats in this proposal?

If not, then I can start on this issue.

koenvo:
Fixed in https://github.com/PySport/kloppy/pull/196

PR #196, opened 2023-06-13
Add GoalkeeperEvent

MKlaasman:
PR related to issue described in https://github.com/PySport/kloppy/issues/138

Added GoalkeeperEvent for:
- **StatsBomb**: including BodyPartQualifiers
- **Opta**: including BodyPartQualifiers
- **WyScout**: Added without BodyPartQualifiers (not present in WyScout V2 & V3)
     a) WyScout V2: Current implementation has qualifiers: SAVE, SAVE_ATTEMPT & REFLEXES (last 2 were also already present for opponent Shots).
     b)WyScout V3: Added SAVE and REFLEXES

Not totally certain about the WyScout implementation, so could use some feedback. 

Let me know whether I should change anything @probberechts & @koenvo .


kloppy/infra/serializers/event/wyscout/deserializer_v2.py
```diff
@@ -103,13 +103,7 @@ def _parse_shot(raw_event: Dict, next_event: Dict) -> Dict:
 
     if next_event["eventId"] == wyscout_events.SAVE.EVENT:
         if next_event["subEventId"] == wyscout_events.SAVE.REFLEXES:
-            qualifiers.append(
-                GoalkeeperActionQualifier(GoalkeeperAction.REFLEX)
-            )
-        if next_event["subEventId"] == wyscout_events.SAVE.SAVE_ATTEMPT:
```
koenvo, line 109: Not sure I understand the removal of the `SAVE_ATTEMPT` here. Does it never happen as it only happens on `wyscout_events.SAVE.EVENT` ?

kloppy/domain/models/event.py
```diff
@@ -344,14 +346,34 @@ class BodyPartQualifier(EnumQualifier):
     value: BodyPart
 
 
-class GoalkeeperAction(Enum):
```
koenvo, line 347: I'm not a huge fan of replacing (renaming) this call. It means a breaking change and could break code that's currently using it. 

For me this is the only blocker to merge it.

probberechts: For consistency with "PassType", "DuelType", etc., I would say that adding "-Type" is definitely more appropriate. Maybe "GoalkeeperActionType" would be the best name here? 

Yet, I agree with Koen that this can't be changed suddenly. What about creating an alias with a deprecation warning?

koenvo: The alias approach sounds good @probberechts combined with `GoalkeeperActionType`

MKlaasman: @koenvo, I could use some clarification on implementing the deprecation warning? 
Does only adding a warning to the docstring suffice? Something like:
"
[DEPRECATED] GoalkeeperAction

   This class is deprecated. Use GoalkeeperActionType instead.
   "

MKlaasman:
Deprecated the `GoalkeeperAction` class by introducing a `DeprecatedEnumValue` class, since the existing @deprecated decorator in utils does not work on Enum Classes.

Using GoalkeeperAction like for instance: GoalkeeperAction.SAVE will now emit the following DeprecationWarning: "DeprecationWarning: GoalkeeperAction is deprecated. Use GoalkeeperActionType instead.
 action = GoalkeeperAction.SAVE"

Not sure whether this is the desired deprecation, feel free to share comments.

@koenvo

PR #197, opened 2023-06-14
Fixed foot left/right typo in shot parsing for WyScout v3

DrSleep:
As per the title, a very minor fix to correctly assign the body part of the shot.

Unfortunately, `wyscout_events_v3.json` does not have any shot events with a defined body part preventing me from adding an appropriate test. If someone could please let me know where to find other events of the match in that json file, I can easily add the test.


JanVanHaaren (approved):
Thanks!

PR #201, opened 2023-06-30
Add speed parsing for SecondSpectrum

DriesDeprest:


kloppy/domain/models/tracking.py
```diff
@@ -24,6 +24,7 @@ class Frame(DataRecord):
     players_data: Dict[Player, PlayerData]
     other_data: Dict[str, Any]
     ball_coordinates: Point3D
+    ball_speed: Optional[int] = None
```
JanVanHaaren, line 27: Is this type hint correct? Based on the test, I'd expect a `float` instead of an `int`.

JanVanHaaren (commented):
Thanks for the pull request!

kloppy/domain/models/tracking.py
DriesDeprest: You are correct it should be float

JanVanHaaren (approved):

Issue #214, opened 2023-08-17
Different start time second half SkillCorner and DataFactory compared to other Providers


UnravelSports:
Looking through [SkillCorner Tests](https://github.com/PySport/kloppy/blob/master/kloppy/tests/test_skillcorner.py) and [DataFactory Tests](https://github.com/PySport/kloppy/blob/master/kloppy/tests/test_datafactory.py) it seems for both of these providers the second half start time is 2700 where as for some other providers (Opta, StatsBomb, Sportec) the second half start time resets to 0. 

If I'm seeing it correctly in [Sportec Tracking Data Tests](https://github.com/PySport/kloppy/blob/master/kloppy/tests/test_sportec.py) we are testing the second period starts at 0. 

```
second_period = dataset.metadata.periods[1]
for frame in dataset:
    if frame.period == second_period:
        assert (
            frame.timestamp == 0
        ), "First frame must start at timestamp 0.0"
        break
else:
    # No data found in second half
    assert False
```

I'm not quite sure if we're doing a similar test for Sportec Event Data.

Now I don't know what the preferred behaviour is, but I would guess the preferred behaviour would be for all of these settings to be the same across providers and event + tracking data, no?

PR #215, opened 2023-08-17
Add SkillCorner extrapolated_data.jsonl loading functionality

UnravelSports:
I noticed there was no support for loading `extrapolated_data.jsonl` files, only deprecated `structured_data.json` files.

As you can see in the messy commit history I've added a couple things:

- Functionality to load `jsonl` files if that file suffix is in place for `raw_data`. This functionality simply converts the json lines to a list of json objects such that it looks exactly like the `structured_data.json` files. We also rename the `timestamp` key to `time` to make it compatible with the rest of the code (that relies on the old `time` key).
- Added improved timestring parsing because the `extrapolated` files have a slightly different timestamp format. The old format is `"92:37.60"` whereas the new format would display this as `"01:32:37.60"`. The `_timestamp_from_timestring()` function can now handle both cases, converting both of these formats to the same output.
- Added an improvement to reset the time at the start of each half to 0. As mentioned in https://github.com/PySport/kloppy/issues/214 this was not implemented yet. This might be a breaking change for some people relying on timestamps, so we might have to think about how to actually implement this last part.

Finally, I would not be surprised if the tests fail because of the change in period start time...


DriesDeprest:
@JanVanHaaren @koenvo what are your thoughts on the last point raised regarding the start timestamp of the second period?

What seems like a good plan to bring this work regarding supporting SkillCorners .jsonl tracking data files into kloppy while considering backwards compatibility?

DriesDeprest:
@koenvo @UnravelSports this [commit](https://github.com/PySport/kloppy/commit/0fd9ad6a1dd116b932717da948999d0a547819ae#diff-2932ebed1fc21704c8b8f5847df9d9fef6196c1c23534746e4ed5e3f0c0b8a95) should still be applied prior to merging imo. This way we use the SkillCorner ID as the player ID in the dataset. I'd love to do it myself but I can't contribute to this PR. So can any of you guys apply the changes?

@koenvo do you know why some of the checks are failing?

JanVanHaaren:
> @koenvo do you know why some of the checks are failing?

The checks should succeed again after merging in `master`. The issue was fixed in #231.

JanVanHaaren:
@UnravelSports Could you please merge the `master` branch into your feature branch? Thanks!

UnravelSports:
It seems I've made a mess out of this, I'm creating a new PR to make a clean PR 😆

Issue #212, opened 2023-08-17
Minor issues in Sportec deserializer


UnravelSports:
Hi, 

Here is a short list of issues I encountered when loading Sportec event data:

- [ ] There appears to be a single event with a negative timestamp. Not sure if this is consistent throughout matches.
- [ ] The kick-off of the second half is assigned as the last event of the first half. I would not be surprised if this is related to the first issue.
- [ ] The [y-coordinates for the event data](https://github.com/PySport/kloppy/blob/master/kloppy/domain/models/common.py#L532) is inverse to the [y-coordinates of the tracking data](https://github.com/PySport/kloppy/blob/master/kloppy/domain/models/common.py#L556) all else being equal.

Issue #211, opened 2023-08-17
Missing "own goals" in Sportec deserializer


UnravelSports:
I noticed the [Sportec Deserializer ](https://github.com/PySport/kloppy/blob/master/kloppy/infra/serializers/event/sportec/deserializer.py) does not support Own Goal events.

For reference, this is what an anonymized Sportec own goal looks like.

```
<Event Y-Source-Position="37.11" Y-Position="34.05" EventId="182094" EventTime="2023-12-17T12:19:56.700+01:00" MatchId="DFL-MAT-A" X-Position="91.90" X-Source-Position="102.20">
 <OwnGoal Player="DFL-OBJ-PLAYER_A" BallPossessionPhase="700" TeamOwnGoalCountsFor="DFL-CLU-TEAM_B" Team="DFL-CLU-TEAM_A" ShotOrigin="3" GoalZone="13" OwnGoalFromCounterAttack="false" TypeOfShot="leftLeg" CurrentResult="3:3" SetupOrigin="left" ScorerBallTouch="direct" ScorerSetup="oppositionGoalkeeperAction" BuildUp="crossOpenPlay"/>
</Event>
```

UnravelSports:
Fixed, PR to follow

probberechts:
This was fixed in https://github.com/PySport/kloppy/pull/217.

PR #216, opened 2023-08-18
Fix Sportec coordinate scaling and own goal issue

UnravelSports:
Partially resolving https://github.com/PySport/kloppy/issues/212 and resolving https://github.com/PySport/kloppy/issues/211

- We add a `transformer.change_point_dimensions()` to the interpreted end coordinates
- We add `OWN_GOAL` as a ShotResult option

Edit: The formatting checks fail, but I've ran `black kloppy/infra/serializers/event/sportec/deserializer.py`... Don't know what that's about


UnravelSports:
I've also added changed Orientation from Orientation.FIXED_HOME_TEAM to Orieantation.HOME/AWAY_TEAM for both Tracking and Event data, because the assumption that the Orientation is fixed is wrong. Since we have different attacking directions for both period 1 and 2 for both event and tracking data sets

PR #217, opened 2023-08-18
Fix sportec current

koenvo:
This PR is based on [#216](https://github.com/PySport/kloppy/pull/216)
- Add `OwnGoal` result
- Use `X/Y-Source-Position` as `receiver_coordinates` when possible

Closes https://github.com/PySport/kloppy/pull/216


kloppy/infra/serializers/event/sportec/deserializer.py
```diff
@@ -584,6 +572,13 @@ def deserialize(self, inputs: SportecEventDataInputs) -> EventDataset:
 
                 events.append(transformer.transform_event(event))
 
+        for i, event in enumerate(events[:-1]):
```
koenvo, line 575: Should use `X-Source-Position`/`Y-Source-Position` here.

koenvo:
@unravelsports can you check this one?

JanVanHaaren (approved):
The pull request looks good as far as I can tell.

Issue #210, opened 2023-08-17
Missing "own goals" in StatsBomb deserializer


UnravelSports:
I noticed the [StatsBomb Deserializer ](https://github.com/PySport/kloppy/blob/master/kloppy/infra/serializers/event/statsbomb/deserializer.py#L588) does not support Own Goal events.

A StatsBomb own goal has two events one "Own Goal For" (id = 25) and one "Own Goal Against" (id = 20). We can probably just use only 1. 

For reference, this is what an anonymized  StatsBomb own goal looks like. 

```
{
    "id" : "f942c5b5-df4b-4ee4-9e90-ed5f5",
    "index" : 16,
    "period" : 1,
    "timestamp" : "00:44:51.456",
    "minute" : 44,
    "second" : 51,
    "type" : {
      "id" : 25,
      "name" : "Own Goal For"
    },
    "possession" : 101,
    "possession_team" : {
      "id" : 1,
      "name" : "Team A"
    },
    "play_pattern" : {
      "id" : 2,
      "name" : "From Corner"
    },
    "obv_for_after" : null,
    "obv_for_before" : null,
    "obv_for_net" : null,
    "obv_against_after" : null,
    "obv_against_before" : null,
    "obv_against_net" : null,
    "obv_total_net" : null,
    "team" : {
      "id" : 101,
      "name" : "Team A"
    },
    "location" : [ 115.7, 39.9 ],
    "duration" : 0.0,
    "related_events" : [ "89dd4f4b-0a70-48d8-a0e7-ac4c" ]
  }, {
    "id" : "89dd4f4b-0a70-48d8-a0e7-ac4c",
    "index" : 17,
    "period" : 1,
    "timestamp" : "00:44:51.456",
    "minute" : 44,
    "second" : 51,
    "type" : {
      "id" : 20,
      "name" : "Own Goal Against"
    },
    "possession" : 101,
    "possession_team" : {
      "id" : 101,
      "name" : "Team A"
    },
    "play_pattern" : {
      "id" : 2,
      "name" : "From Corner"
    },
    "obv_for_after" : null,
    "obv_for_before" : null,
    "obv_for_net" : null,
    "obv_against_after" : null,
    "obv_against_before" : null,
    "obv_against_net" : null,
    "obv_total_net" : null,
    "team" : {
      "id" : 102,
      "name" : "Team B"
    },
    "player" : {
      "id" : 10000,
      "name" : "Player Who Scored Own Goal"
    },
    "position" : {
[...]

PR #219, opened 2023-08-26
Add owngoal event for SB

AlienDeg:
https://github.com/PySport/kloppy/issues/210


koenvo:
Hi! Thanks for this PR. 

Just did a quick check of the code. 

We already have a way to define an Own Goal: https://github.com/PySport/kloppy/blob/master/kloppy/domain/models/event.py#L60

The Statsbomb deserializer should use a `ShotEvent` with result `ShotResult.OWN_GOAL`.

Issue #226, opened 2023-10-19
Metrica EPTS loader fails when Score field is not contained in the metadata file


jcnunezmoreno:
Metrica EPTS loader try to load Score field from the metadata xml file, which is expected to be on Sessions/MatchParameters

jcnunezmoreno:
Hi @koenvo,
Do you have a plan for a new release? Just to know if we can wait for it or if we need to integrate our own fork. 
Thanks in advance!

PR #227, opened 2023-10-19
Fixing epts metrica loader when score is not on the metadata file

jcnunezmoreno:
This PR fix issue #226: Metrica EPTS loader fails when Score field is not contained in the metadata file


kloppy/infra/serializers/tracking/metrica_epts/metadata.py
```diff
@@ -227,21 +227,29 @@ def load_metadata(
...
+            home=score_elm.LocalTeamScore, away=score_elm.VisitingTeamScore
+        )
+
+        _team_map = {
+            score_elm.attrib["idLocalTeam"]: Ground.HOME,
+            score_elm.attrib["idVisitingTeam"]: Ground.AWAY,
+        }
+    else:
+        score = Score(home=0, away=0)
+        _team_map = {
```
koenvo, line 248: Could you replace this with something like
`home_team_id, away_team_id = list(_team_name_map.keys())`

koenvo: And does it makes sense to always use this logic instead of the 
```python
        _team_map = {
            score_elm.attrib["idLocalTeam"]: Ground.HOME,
            score_elm.attrib["idVisitingTeam"]: Ground.AWAY,
        }
```
code?

koenvo: Okee that makes sense. Thanks.

Could you change the `home_team_id, away_team_id = list(_team_name_map.keys())` part? Than I'm able to merge the PR.

jcnunezmoreno: Done! :)

PR #230, opened 2023-11-10
Add InterceptionEvent

### LINKED ISSUE #222: Add a data model for interceptions
Kloppy's event data model currently does not support interceptions. These are commonly defined as "an attempt by a defending player to intercept a pass between opposition players to prevent the ball from reaching its target.".

### How data providers do it
**StatsBomb**, **Wyscout** and **Opta** all have an "Interception" event. I think **Wyscout** and **Opta** only record successful interceptions (check this; the docs are unclear). **StatsBomb** also records failed interceptions. When an interception is immediately followed by a pass, **StatsBomb** and **Wyscout** do not record an "Interception" event and the pass gets typed/tagged as an “Interception”. Maybe, these passes have to be split into two events again when doing the conversion to the Kloppy format.

### Proposal
Add an `InterceptionEvent` with an `InterceptionResult`. As possible values for the result, I would suggest:
- Success: An interception that gains possession of the ball (without going out of bounds)
- Lost: An interception by the defending team that knocked the ball to an attacker
- Out: An interception that knocked the ball out of bounds

Maybe for Wyscout and Opta, the result can be inferred from the next action.

Passes that are labeled as an interception should by split into a "Interception" and "Pass" event.

MKlaasman:
fixes #222 

Notes:
- Whenever needed (wyscout & opta), I inferred the InterceptionResult from the next event. Though, this approach might still be suboptimal.
- In WyScout v2: The orginal RecoveryEvent gets overwritten by the InterceptionEvent, whenever possible

What do you think? @probberechts  @JanVanHaaren

### DIFF
--- a/kloppy/infra/serializers/event/opta/deserializer.py
+++ b/kloppy/infra/serializers/event/opta/deserializer.py
@@ -1,4 +1,4 @@
-from typing import Tuple, Dict, List, NamedTuple, IO
+from typing import Tuple, Dict, List, NamedTuple, IO, Optional
 import logging
 from datetime import datetime
 import pytz
@@ -78,6 +78,7 @@ EVENT_TYPE_CARD = 17
 EVENT_TYPE_RECOVERY = 49
 EVENT_TYPE_FORMATION_CHANGE = 40
 EVENT_TYPE_BALL_TOUCH = 61
+EVENT_TYPE_BLOCKED_PASS = 74
 
 EVENT_TYPE_SAVE = 10
 EVENT_TYPE_CLAIM = 11
@@ -391,7 +392,7 @@ def _parse_interception(
     qualifiers = _get_event_qualifiers(raw_qualifiers)
     result = InterceptionResult.SUCCESS
 
-    if next_event:
+    if next_event is not None:
         next_event_type_id = int(next_event.attrib["type_id"])
         if next_event_type_id in BALL_OUT_EVENTS:
             result = InterceptionResult.OUT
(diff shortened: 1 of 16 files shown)


kloppy/infra/serializers/event/opta/deserializer.py
```diff
@@ -761,6 +792,14 @@ def deserialize(self, inputs: OptaInputs) -> EventDataset:
                             **duel_event_kwargs,
                             **generic_event_kwargs,
                         )
+                    elif type_id == EVENT_TYPE_INTERCEPTION:
```
probberechts, line 795: Opta also has a "blocked pass" event. Would it make sense to parse it as an interception too?

blocked pass := When a player tries to cut out an opposition pass by any means. Similar to an interception except there is much less reading of the pass.

kloppy/infra/serializers/event/wyscout/deserializer_v2.py
```diff
@@ -477,6 +510,24 @@ def deserialize(self, inputs: WyscoutInputs) -> EventDataset:
                         **generic_event_args,
                     )
 
+                # If also an interception event, add this before other event (It is a tag on multiple wyscout_events)
```
probberechts, line 513: Wyscout does not have a separate event type for interceptions. Instead, interceptions are recorded as a pass, duel or touch event with an "interception" tag. Your current implementation works well for passes, but for consistency with other providers (and socceraction) I would convert these duels/touch events to an interception. Not to an interception + duel.

probberechts (commented):
I've only added two suggestions. Apart from these it seems ok to me. Thanks for implementing this!

Note that I did not check the Wyscout v3 implementation in detail since I am not familiar with that data.

MKlaasman:
Addition:
1) Wyscout v2: Now duels and touches are overwritten when they have an interception tag. 
2) Opta: Handle a blocked pass as an interception.
3) Opta: Filtered deleted events upfront, otherwise next_event implementation prone for wrong InterceptionResult. Now, not logging the Skipping of events anymore. Someone, thinking this is a bad design choice?

probberechts (approved):

JanVanHaaren (approved):

koenvo:
@MKlaasman thanks for your work. Can you please merge master in? There are some small merge conflicts due to changes from Pieters PRs.

PR #238, opened 2023-11-29
Recognize take on event for Wyscout

DriesDeprest:


kloppy/infra/serializers/event/wyscout/deserializer_v3.py
```diff
@@ -299,6 +299,28 @@ def _parse_set_piece(raw_event: Dict, next_event: Dict, team: Team) -> Dict:
     return result
 
 
+def _parse_takeon(raw_event: Dict) -> Dict:
```
JanVanHaaren, line 302: I would rename this function to `_parse_take_on` for consistency with `event_factory.build_take_on`.

JanVanHaaren (approved):
Thanks!

PR #243, opened 2023-12-01
Add current formation and opponent formation info to event, start with parser for Opta

DriesDeprest:
I want to add current formation and current opponent formation information to events. This would allow us to more easily filter out events of a team or player when they were playing in a certain formation or against a certain formation. E.g. only analyse events where Lionel Messi was playing as a LW in a 4-4-2 against a 4-3-3.

### DIFF
--- a/kloppy/domain/models/event.py
+++ b/kloppy/domain/models/event.py
@@ -506,6 +506,8 @@ class Event(DataRecord, ABC):
     raw_event: Dict
     state: Dict[str, Any]
     related_event_ids: List[str]
+    formation: Optional[FormationType]
+    opponent_formation: Optional[FormationType]
 
     qualifiers: List[Qualifier]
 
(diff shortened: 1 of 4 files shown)


DriesDeprest:
@koenvo @JanVanHaaren sorry to bother, but I'm failing to understand why I'm getting this error: 
TypeError: GenericEvent.__init__() missing 2 required positional arguments: 'formation' and 'opponent_formation'

Why is the error message saying 'formation' and 'opponent_formation' are required, while I label them as Optional in my class?

```
formation: Optional[FormationType]
opponent_formation: Optional[FormationType]
```

koenvo:
Thanks for the issue Dries!

In this case the state builder would be a good fit. Have you seen https://kloppy.pysport.org/examples/state/#Add-state---lineup ? Would this work for your case?

DriesDeprest:
I forgot about the state builder. Yes, indeed, the state builder solves my use case.

However, Wyscout doesn't have a formation change event, it rather encapsulates the formation and opponent formation in each event, see: https://github.com/PySport/kloppy/blob/master/kloppy/tests/files/wyscout_events_v3.json#L57-L127. How would you suggest that we pass the current (opponent) formation information to the events for this data provider?

Would you save the formation information of the prior event in the Wyscout deserializer and check for a difference, and if so create an "artificial" formation change event?

koenvo:
The "artificial" formation change event sounds like a great solution.

PR #246, opened 2023-12-03
Estimate Wyscout v2 shot result coordinates

probberechts:
The result coordinates of shots in Wyscout v2 data are always (0,0) and instead tags such as "Goal low center" are used to indicate where the shot ended. With this commit, the coordinates are estimated from these tags.

### DIFF
--- a/kloppy/infra/serializers/event/wyscout/deserializer_v2.py
+++ b/kloppy/infra/serializers/event/wyscout/deserializer_v2.py
@@ -1,6 +1,6 @@
 import json
 import logging
-from typing import Dict, List, Tuple, NamedTuple, IO
+from typing import Dict, List, Tuple, NamedTuple, IO, Optional
 
 from kloppy.domain import (
     BodyPart,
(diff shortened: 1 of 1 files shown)


kloppy/infra/serializers/event/wyscout/deserializer_v2.py
```diff
@@ -80,6 +80,62 @@ def _generic_qualifiers(raw_event: Dict) -> List[Qualifier]:
...
+        or _has_tag(raw_event, wyscout_tags.OUT_CENTER_LEFT)
+        or _has_tag(raw_event, wyscout_tags.OUT_HIGH_LEFT)
+    ):
+        return Point(100.0, 40.0)
+    if (
+        _has_tag(raw_event, wyscout_tags.POST_LOW_LEFT)
+        or _has_tag(raw_event, wyscout_tags.POST_CENTER_LEFT)
+        or _has_tag(raw_event, wyscout_tags.POST_HIGH_LEFT)
+    ):
+        return Point(100.0, 55.38)
```
JanVanHaaren, line 124: Would it make sense to store the coordinates in constants that can be re-used throughout the deserializer?

For future contributors, it might perhaps not be obvious that 44.62 and 55.38 are the y-coordinates for the goal posts.

kloppy/infra/serializers/event/wyscout/deserializer_v2.py
```diff
@@ -80,6 +80,62 @@ def _generic_qualifiers(raw_event: Dict) -> List[Qualifier]:
...
+    if (
+        _has_tag(raw_event, wyscout_tags.POST_LOW_RIGHT)
+        or _has_tag(raw_event, wyscout_tags.POST_CENTER_RIGHT)
+        or _has_tag(raw_event, wyscout_tags.POST_HIGH_RIGHT)
+    ):
+        return Point(100.0, 44.62)
+    if _has_tag(raw_event, wyscout_tags.BLOCKED):
+        return Point(
```
JanVanHaaren, line 132: Do I understand correctly that the start location is also used as the end location when a shot is blocked because the actual location of the block can't be inferred from the data?

JanVanHaaren (commented):
Thanks! Would it be possible to also add a test for this functionality?

kloppy/infra/serializers/event/wyscout/deserializer_v2.py
probberechts: Yes, I followed the implementation in socceraction. I guess it's the best possible estimate of the shot's end location. Another option would be to set it to `None`. Do you have a preference?

JanVanHaaren: Thanks! I think this approach is fine. I just wanted to check whether it was a deliberate choice or a potential mistake.

kloppy/infra/serializers/event/wyscout/deserializer_v2.py
probberechts: These coordinates are not used anywhere else, so I don't really see the value of adding more constants. I've added some documentation instead. I hope that's ok too.

JanVanHaaren: Thanks. Looks good to me!

JanVanHaaren (approved):

PR #250, opened 2023-12-04
Add artificial formation change event in Wyscout deserializer

DriesDeprest:
Follow-up on https://github.com/PySport/kloppy/pull/243


kloppy/infra/serializers/event/wyscout/deserializer_v3.py
```diff
@@ -547,12 +616,40 @@ def deserialize(self, inputs: WyscoutInputs) -> EventDataset:
...
+                            raw_event, next_event, teams, home_team, away_team
+                        )
+                    )
+                    for (
+                        formation_change_team,
+                        formation_change_event_kwargs,
+                    ) in event_formation_change_info.items():
+                        generic_event_args.update(
+                            {
+                                "event_id": None,
```
koenvo, line 637: It would be good to give this event an id. I checked the Metrica implementation which also creates synthetic events and it seems it's reusing the original event id. 

The eventIds should be unique over a match. Maybe you can add a prefix to the original eventId and use that one? 

And would be good to use the term 'synthetic' instead of 'artificial'.

JanVanHaaren (approved):

koenvo:
Thanks!

I merged all aproved PR's so far. This caused a merge conflict in this PR. Can you please merge master in this branch? When the conflict is resolved I'll merge it.

PR #259, opened 2023-12-15
Fix receiver_coordinates of passes for Wyscout v2

probberechts:
Wyscout v2 does not define the end coordinates of blocked passes: `positions[1]` defaults to (0,0). This commit fills in the receiver_coordinates attribute with the start location. This is a good approximation since passes are by definition blocked near their start location.


probberechts:
As a side note: `receiver_coordinates` might not be the best name. This attribute is actually filled in with the result coordinates / end coordinates of the pass in (almost?) all serializers, not with the location of the intended receiver.

JanVanHaaren (approved):
Thanks!

koenvo:
Agree on the possibility to improve the the `receiver_coordinates` name to reflect what it is.

PR #260, opened 2023-12-15
Fix GoalKeeperActionType.SAVE coordinates for Wyscout v2

probberechts:
Wyscout v2 does not store the coordinates of the "Save attempt" event as regular start coordinates. Instead, they should be inferred from the end coordinates and be inverted.

For example, in
```
  'positions': [{'y': 100, 'x': 100}, {'y': 66, 'x': 20}]
```
the start location is x=100-66, y=100-20.


JanVanHaaren (approved):
Thanks!

Issue #256, opened 2023-12-13
Add support for parsing events that occur during extra time


DriesDeprest:
I wanted to add support for parsing events that occur during extra time. I believe currently we don't do this for any serializer, as we only consider two possible [Periods](https://github.com/PySport/kloppy/blob/master/kloppy/domain/models/common.py#L334), and as a result events during extra time are skipped (see e.g. [Opta](https://github.com/PySport/kloppy/blob/master/kloppy/infra/serializers/event/opta/deserializer.py#L711-L718)).

Thoughts @koenvo @JanVanHaaren ?

JanVanHaaren:
I like this proposal. I think we could even extend this proposal from extra time to penalty shoot-outs.

dvilches:
Hi, in  [Wyscout v3](https://github.com/PySport/kloppy/blob/897e9d0ee7b58bdbe3b64c99583c99c5c9845d0e/kloppy/infra/serializers/event/wyscout/deserializer_v3.py#L430C11-L430C75) we had the same problem.

period_id = int(raw_event["matchPeriod"].replace("H", ""))

We changed it to:

if "H" in raw_event["matchPeriod"]:
    period_id = int(raw_event["matchPeriod"].replace("H", ""))
else:
    period_id = int(raw_event["matchPeriod"].replace("E", "")) + 2

This solution may not be the most elegant but it works for us.

PR #263, opened 2023-12-20
Add Opta support for extra time & penalty shootouts

DriesDeprest:
PR related to https://github.com/PySport/kloppy/issues/256

### DIFF
--- a/kloppy/infra/serializers/event/opta/deserializer.py
+++ b/kloppy/infra/serializers/event/opta/deserializer.py
@@ -263,6 +263,58 @@ def _parse_f24_datetime(dt_str: str) -> float:
     )
 
 
+def _create_periods(match_result_type: str) -> List[Period]:
+    periods = [
+        Period(
+            id=1,
+            start_timestamp=None,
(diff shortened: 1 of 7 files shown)


kloppy/infra/serializers/event/opta/deserializer.py
```diff
@@ -263,6 +263,58 @@ def _parse_f24_datetime(dt_str: str) -> float:
     )
 
 
+def _create_periods(match_result_type: str) -> List[Period]:
```
JanVanHaaren, line 266: I would use a `for` loop to instantiate the `Period` objects.

```
if match_result_type == "AfterExtraTime":
    num_periods = 4
elif match_result_type == "PenaltyShootout":
    num_periods = 5
else:
    num_periods = 2

periods = [
    Period(
        id=period_id,
        start_timestamp=None,
        end_timestamp=None,
    )
    for period_id in range(1, num_periods + 1)
]
```

JanVanHaaren (approved):
Thanks!

PR #264, opened 2023-12-26
Remove duplicate Metrica set-piece events

probberechts:
Metrica encodes set-pieces as two events. The first event defines the set-piece type (e.g., kick-off) and the second event defines the action type (e.g., pass). In Kloppy, these two events should be integrated in a single event.

Furthermore, I few small fixes and some refactoring:
- Fixes a few type annotations
- Removes unused imports
- Refactor unit tests
- Metrica does not define an event id. The `index` is now used as the event id.
- The `previous_event`  variable wrapped around to the last event of the dataset, which could cause unexpected behavior.


JanVanHaaren (approved):

PR #266, opened 2023-12-26
Sync event data to tracking data

### LINKED ISSUE #61: Adding a way of matching the event with a moment in time within the tracking data
There is a need to match the time of an event with the time within the tracking data, I am trying to implement a way this can be done accurately to remove the error syncing the datasets (opta,tracab). The first step to this is to add a very clumsy sync via the use of timestamps before things are refined. 

I would propose two attributes are added to the event object structure, clumsy_sync and precise_sync which are empty and only filled if the user wishes (as it relies on event data and tracking data to be present. 

filled by the following function.

Interested to hear thoughts 

image

probberechts:
There is still quite a lot of work to be done, but I am already creating this PR to get some initial feedback on how to integrate this into Kloppy. I've created a notebook in `docs/examples/event_tracking_sync.ipynb` that illustrates how to use the code. Apart from a few edge cases in the Needleman-Wunsch alignment method, everything seems to work.

I've now implemented it such that it can be configured as [ETSY](https://github.com/ML-KULeuven/ETSY), [sync.soccer](https://github.com/huffyhenry/sync.soccer), [the method used by Anzer and Bauer](https://www.frontiersin.org/articles/10.3389/fspor.2021.624475/full#h4), or any combination of these. For example, you can use the Needleman-Wunsch alignment method of syn.soccer with the rules of ETSY. This gives the user a lot of flexibility, but I think it would be good to agree on a default method that works well on most datasets. If people would be interested in playing with it and have suggestions for default parameter settings that work well, this would be useful too. 

Fixes #61

**Tentative TODO list:**
- [ ] Add tests
- [ ] Add `EventTrackingSynchronizer.fit(...)` method to tune weights of scoring rules on an annotated dataset
- [ ] Add `EventDataset.sync(tracking_dataset, ...)` method
- [ ] Determine good default configuration
- [ ] Fix Needleman-Wunsch edge cases
- [ ] Optimize speed

----
https://github.com/PySport/kloppy/assets/2175271/87c93c7c-99f1-4288-a000-b0338977b1bc

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -3,6 +3,40 @@
 
 Find out all changes between different kloppy versions
 
+## 3.14.0 (2023-12-29)
+
+Pull requests merged:
+
+- Add SetPieceQualifiers for StatsBomb Shot event ([#223](https://github.com/PySport/kloppy/pull/223))
+- Add CardQualifiers for StatsBomb FoulCommited event ([#221](https://github.com/PySport/kloppy/pull/221))
+- Run build docs ([#231](https://github.com/PySport/kloppy/pull/231))
+- Fixing epts metrica loader when score is not on the metadata file ([#227](https://github.com/PySport/kloppy/pull/227))
+- Fix Bodypart qualifiers in StatsBomb clearance event ([#228](https://github.com/PySport/kloppy/pull/228))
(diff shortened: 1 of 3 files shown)


probberechts:
I've completely messed up my git history 😣 I'll create a new PR later.

PR #278, opened 2024-01-11
Redo of #PR215

UnravelSports:
Because I made a mess of PR https://github.com/PySport/kloppy/pull/215, here is a clean attempt. 
It includes updated jsonl loader without with open(), Dries' minor changes to IDs as proposed in https://github.com/PySport/kloppy/commit/277a7143566df9f05cef5b1d66a353cadcb0f4ba and updated event_factory.

### DIFF
--- a/kloppy/domain/services/event_factory.py
+++ b/kloppy/domain/services/event_factory.py
@@ -41,10 +41,6 @@ def create_event(event_cls: Type[T], **kwargs) -> T:
        but we filter those arguments out.
     """
     extra_kwargs = {"state": {}}
-
-    if "qualifiers" not in kwargs:
-        kwargs["qualifiers"] = None
-
     if "related_event_ids" not in kwargs:
         extra_kwargs["related_event_ids"] = []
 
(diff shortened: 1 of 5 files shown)


JanVanHaaren:
Thanks, @UnravelSports! The changes to the SkillCorner deserializer look good to me.

Could you please fix the [code formatting](https://kloppy.pysport.org/contributing/#code-formatting)? I suspect that the formatting changes to the event data deserializers were unintentional.

JanVanHaaren:
@UnravelSports I've managed to make the necessary changes myself. 😊

JanVanHaaren (approved):

UnravelSports:
@JanVanHaaren for future reference, what formatting was wrong? Because I did run black on all the files, so kinda curious where it went wrong

Thanks for fixing it and approving of the changes!

JanVanHaaren:
> @JanVanHaaren for future reference, what formatting was wrong? Because I did run black on all the files, so kinda curious where it went wrong

Good question. I suspect that you're using a different version of `black` that has slightly different default preferences, but I'm  not entirely sure.

UnravelSports:
Good point, I'll update :)

Issue #236, opened 2023-11-19
Wyscout v2 does not seem to use BALL_OWNING_TEAM orientation


probberechts:
It seems Wyscout v2 uses the `ACTION_EXECUTING_TEAM` orientation. Not the `BALL_OWNING_TEAM` orientation as it is currently implemented.

Below is some proof from the public data (match id 2058007). It is a duel between a player of the attacking team and a player of the defending team. With the `ACTION_EXECUTING_TEAM` orientation, both events should have the same coordinates, but they are mirrored.

```json
      {
         "eventId":1,
         "subEventName":"Ground attacking duel",
         "tags":[
            {
               "id":502
            },
            {
               "id":703
            },
            {
               "id":1801
            }
         ],
         "playerId":14816,
         "positions":[
            {
               "y":65,
               "x":78
            },
            {
               "y":57,
               "x":79
            }
         ],
         "matchId":2058007,
         "eventName":"Duel",
         "teamId":12913,
         "matchPeriod":"1H",
         "eventSec":54.589237000000026,
         "subEventId":11,
         "id":261445464
      },
      {
         "eventId":1,
         "subEventName":"Ground defending duel",
         "tags":[
            {
               "id":501
            },
            {
               "id":701
            },
            {
               "id":1802
            }
         ],
         "playerId":48,
         "positions":[
            {
               "y":35,
               "x":22
            },
            {
               "y":43,
               "x":21
            }
         ],
         "matchId":2058007,
         "eventName":"Duel",
         "teamId":5629,
         "matchPeriod":"1H",
         "eventSec":54.84165100000001,
         "subEventId":12,
         "id":261445573
      },
```

I am not entirely sure though since I don't have access to recent Wyscout data and according to the API docs, it is the "position in percentage from the left corner of the attacking team". 

Anyway, if Wyscout would indeed use the `BALL_OWNING_TEAM` orientation, the serializer should set the ball owning team (which is not defined in the raw data of v2, btw).

I couldn't check it for the v3 data. But that datastream indeed defines the ball owning team, so it might be correct.

DriesDeprest:
No, in Wyscout v3 raw data, the orientation seems to also be `ACTION_EXECUTING_TEAM`. As you can see below. This should thus be fixed, as currently it is implemented as `BALL_OWNING_TEAM`.
```json
{
      "id": 1,
      "matchId": 1,
      "matchPeriod": "1H",
      "minute": 3,
      "second": 26,
[...]

PR #279, opened 2024-01-12
Fix Wyscout orientation

DriesDeprest:
Wyscout v2 and v3 have` ACTION_EXECUTING_TEAM` as orientation. See issue https://github.com/PySport/kloppy/issues/236 for context.


JanVanHaaren:
Thanks! Would it be possible to add some tests to prevent this functionality from breaking in the future?

JanVanHaaren (approved):
Thank you!

Issue #239, opened 2023-11-30
PassType.ASSIST is being used both as a shot and goal assist interchangeably


DriesDeprest:
In our Opta deserializer, we add the `PassType.ASSIST` qualifier to a pass with an Opta qualifier type 210, see:
https://github.com/PySport/kloppy/blob/master/kloppy/infra/serializers/event/opta/deserializer.py#L543-L544
qualifier type 210 means a shot assist, as you can see below:
<img width="806" alt="image" src="https://github.com/PySport/kloppy/assets/46134788/50cb1739-0b9e-4f35-b7f7-6ec907c760c6">

In our Statsbomb deserializer, we add the `PassType.ASSIST` qualifier to a pass with a "goal_assist" tag, see:
https://github.com/PySport/kloppy/blob/master/kloppy/infra/serializers/event/statsbomb/deserializer.py#L291-L293

I would suggest moving from `PassType.ASSIST` to a `PassType.SHOT_ASSIST`and `PassType.GOAL_ASSIST`. A goal assist would have both qualifiers. This way we are explicit about what we mean and we are capable of handling both shot and goal assists.

@koenvo agree?

DriesDeprest:
I think your definition of assist being a pass being followed directly by a shot event is too narrow.

If you look at the raw data of vendors (e.g. StatsBomb) you see that in between the pass, which gets annotated as an assist, and the shot, there often are carry or duel events, maybe even others.

Therefore, I would use the qualifiers in the raw data where possible. For vendors which don't support shot or goal assists, you can always still try to calculate it.

JanVanHaaren:
I agree that we should make the use of the `PassType.ASSIST` qualifier consistent across the deserializers for the different data providers. This example nicely illustrates that we need more formal definitions for our events and qualifiers to avoid ambiguity as data providers sometimes use the same terms for different concepts.

I'm also in favor of deriving qualifiers, and in some cases even events, from the raw data as much as possible. I believe that this approach would help us to arrive at more uniform and predictable behavior across our deserializers, especially for rather well-defined concepts such as shot assists or key passes and goal assists. However, in other cases, the qualifiers that data providers add to events might not be reconstitutable from the context.

[...]

PR #281, opened 2024-01-16
Add shot assist to data model and fix Opta & StatsBomb deserializers accordingly

DriesDeprest:
PR related to https://github.com/PySport/kloppy/issues/239.

I've added `SHOT_ASSIST` as a possible`PassType` to our data model and fixed the Opta & StatsBomb deserializers accordingly. We now correctly make a distinction between a shot assist and (goal) assist.

I was in doubt on whether we wanted to rename `ASSIST` to `GOAL_ASSIST` to be more explicit in our naming, but didn't do it as I didn't want to introduce breaking changes.

I also chose to add both a `SHOT_ASSIST` and `ASSIST` qualifier to a goal assist, you could argue to only add `ASSIST` qualifier to a goal assist. I don't have a strong preference on that.


DriesDeprest:
@JanVanHaaren @koenvo any updates on this PR? Wondering what your thoughts are on this.

kloppy/domain/models/event.py
```diff
@@ -350,7 +350,9 @@ class PassType(Enum):
     THROUGH_BALL = "THROUGH_BALL"
     CHIPPED_PASS = "CHIPPED_PASS"
     FLICK_ON = "FLICK_ON"
+    SHOT_ASSIST = "SHOT_ASSIST"
     ASSIST = "ASSIST"
+    SHOT_ASSIST_2ND = "SHOT_ASSIST_2ND"
```
JanVanHaaren, line 355: What is the purpose of `SHOT_ASSIST_2ND`?

DriesDeprest: That is a mistake. My original idea was to also add create pre-assist qualifiers, but think we should keep that out of scope of this PR. I'll fix it.

DriesDeprest: I added it here as a sort of placeholder to make the data model more "complete", so that we have all possible options related to pre/normal and shot/goal assists: `SHOT_ASSIST` `ASSIST` `SHOT_ASSIST_2ND` `ASSIST_2ND`.

I don't think StatsPerform, StatsBomb or Wyscout has qualifiers for `SHOT_ASSIST_2ND`, but I can image this changing or there being other (future) providers that do add qualifiers for this.

But if you prefer to remove it from the data model as long as we don't have a data provider supporting it, happy to remove it.

JanVanHaaren (approved):

PR #287, opened 2024-01-20
[Opta] Add BALL_STATE flag

probberechts:
The Opta deserializer effectively sets the "ball_state" attribute for each event, but the corresponding flag was not set in the metadata.


JanVanHaaren (approved):
Thanks!

Issue #267, opened 2023-12-27
Ordering of Opta events


probberechts:
I noticed that Opta events can sometimes be slightly out of order. The F24 docs specify that the following attributes (in the given order) should be used to order each team's match events chronologically:

![image](https://github.com/PySport/kloppy/assets/2175271/b188f5ae-8c49-46b1-a8a9-9be329336f25)

Only sorting by timestamp does not always give the same result. For example:
```xml
<Event id="1889768843" event_id="358" type_id="1" period_id="1" min="32" sec="3" player_id="59062" team_id="174" outcome="0" x="21.6" y="39.2" timestamp="2018-08-20T21:32:27.98" last_modified="2018-08-20T21:32:28" version="1534797148460"></Event>         
<Event id="1592827425" event_id="228" type_id="1" period_id="1" min="32" sec="4" player_id="80908" team_id="957" outcome="0" x="60.4" y="52.0" timestamp="2018-08-20T21:32:27.635" last_modified="2018-08-21T16:43:18" version="1534866198424"></Event>
```

Since the Opta deserializer currently only parses the "timestamp" field, it does not seem possible to order events chronologically.

koenvo:
Are there any details on how to properly sort on correctly and maintain millisecond precision?

A solution could be to extract timestamp from “min” and “sec” attributes but than we lose the precision.

probberechts:
My documentation doesn't mention the precision of the "timestamp" field. However, my version of the documentation is extremely outdated. Maybe @JanVanHaaren has something more up-to-date.

I find it strange that the "timestamp" field does not align with the "min" and "sec" fields. If the precision of the "timestamp" field would be inferior to the "min" and "sec" fields, I don't see why we would infer an (incorrect) millisecond precision from it.

probberechts:
Looking at a few more timestamps, I now realize that Opta does not add leading zeros to the milliseconds. So, "2018-08-20T21:32:27.98" is actually "2018-08-20T21:32:27.000098". 

Python's `%f` pads zeros to the right,  while we should pad zeros to the left to parse the Opta timestamp. We should simply adapt the timestamp parser and then it should work.

> %f is an extension to the set of format characters in the C standard (but implemented separately in datetime objects, and therefore always available). When used with the strptime() method, the %f directive accepts from one to six digits and zero pads on the right.

JanVanHaaren:
[...]

PR #288, opened 2024-01-20
[Opta] Fix F24 timestamp parsing

probberechts:
Opta does not zero-pad milliseconds. Therefore, they were incorrectly parsed by Python's default "%f" format code.

See also #267


JanVanHaaren (approved):
Thanks!

PR #290, opened 2024-01-20
[Metrica] Fix metadata after transform

probberechts:
The pitch dimensions and coordinate system were not updated when deserializing metrica data with the non-default coordinate system.


JanVanHaaren (approved):
Thanks!

Issue #284, opened 2024-01-19
Add counter attack parsing in StatsBomb deserializer


DriesDeprest:
I would like to recognize the "From Counter" play pattern in our StatsBomb deserializer and thus add `CounterAttackQualifier` s to events with this play pattern.

I'm not sure what would be an elegant way to implement this.

I could create a function `_get_counter_attack_qualifiers(event_dict: Dict) -> List[CounterAttackQualifier]` which recognizes the "From Counter" play pattern. And then add this function in all the `_create_events()` methods of the different event type classes. But it feels a bit repetitive to add it to almost all of these event classes - @probberechts thoughts?

probberechts:
Ah yes, that would indeed be repetitive.

You could add the qualifier after parsing the events:

```python
    def deserialize(self, event_factory: EventFactory) -> List[Event]:
        """Deserialize the event.
        """
        generic_event_kwargs = self._parse_generic_kwargs()
        events = (
            self._create_aerial_won_event(
                event_factory, **generic_event_kwargs
            )
            + self._create_events(event_factory, **generic_event_kwargs)
            + self._create_ball_out_event(
                event_factory, **generic_event_kwargs
            )
        )
        for event in events:
            # add possession qualfiers
        return events
```

PR #291, opened 2024-01-22
identify counter attacks for StatsBomb

DriesDeprest:
Related to https://github.com/PySport/kloppy/issues/284.

(Made a mistake for the commit message, you can ignore that.)


kloppy/tests/test_statsbomb.py
```diff
@@ -863,6 +864,10 @@ def test_loose_ground_duel_qualfiers(self, dataset: EventDataset):
             DuelType.GROUND,
         ]
 
+    def test_counter_attack_qualifier(self, dataset: EventDataset):
```
JanVanHaaren, line 867: I think it would be good if we added a few more tests. For instance, check the number of counter-attacking events and check whether a non-counter-attacking event doesn't have the qualifier.

JanVanHaaren:
@probberechts Do you have any feedback on this pull request? Thanks.

DriesDeprest:
@koenvo, this should be ready to merge!

PR #294, opened 2024-01-27
Uniform implementation of timestamps

probberechts:
The representation of timestamps was inconsistent between different data providers.

This commit implements a uniform implementation between data providers. The **timestamps of events and frames are now relative to the kick-off event in each period**. This implies that timestamps are reset to zero for each period and that tracking frames before the kick-off event get a negative timestamp. The standard implementation for the start_timestamp and end_timestamp of periods is the absolute UTC datetime of respectively the kick-off and final whistle of the corresponding period. If the absolute datetime is not available, we use the offset between the start of the period's data feed and the kick-off for the start_timestamp, and the offset between the start of the period's data feed and the final whistle for the end_timestamp.

## BREAKING CHANGES: 
- The `Period.start_timestamp` and `Period.end_timestamp` fields now have type datetime or timedelta instead of float
- The `DataRecord.timestamp` field now has type timedelta instead of float
- The value of the `Period.start_timestamp`, `Period.end_timestamp` and `DataRecord.timestamp` fields changes for most data providers


probberechts:
For Wyscout v3 it was unclear whether timestamps are reset at the start of each period. That is, does the first event of the second period get timestamp "00:00" or "45:00". This is not clear from the test data which contains artificial data. Maybe @DriesDeprest can answer this?

JanVanHaaren:
> For Wyscout v3 it was unclear whether timestamps are reset at the start of each period. That is, does the first event of the second period get timestamp "00:00" or "45:00". This is not clear from the test data which contains artificial data.

In Wyscout V3 event data, the second period starts at timestamp "45:00".

```json
{
    "matchPeriod": "2H",
    "minute": 45,
    "second": 2,
    "matchTimestamp": "00:45:02.581",
    "videoTimestamp": "3011.581222",
}
```

probberechts:
Thanks, @JanVanHaaren! I guessed incorrectly 😅 but it should be fixed.

kloppy/domain/models/common.py
```diff
@@ -238,21 +240,32 @@ class Period:
...
 
-    def contains(self, timestamp: float):
-        return self.start_timestamp <= timestamp <= self.end_timestamp
+    def contains(self, timestamp: datetime):
+        if isinstance(self.start_timestamp, datetime) and isinstance(
+            self.end_timestamp, datetime
+        ):
+            return self.start_timestamp <= timestamp <= self.end_timestamp
+        raise KloppyError(
+            "This method can only be used when start_timestamp and end_timmestamp are a datetime"
```
JanVanHaaren, line 264: Minor: `end_timmestamp` -> `end_timestamp`.

JanVanHaaren (approved):
Thank you for this nice contribution! This pull request looks good to me, besides one small typing mistake in an error message.

PR #295, opened 2024-01-29
Add support for JSON Tracab data

DriesDeprest:
**Goal**
Add support to serialize tracking data from Tracab, where both the meta and the raw data are in JSON format. 

**Description**
Refactored the code, so we can both serialize tracking data from Tracab, if a) the meta data is XML and raw data .dat or b) the meta data is JSON and raw data JSON. 

I've refactored the code to support two versions of input data similar to how we handle it for the Wyscout v2 and v3 event data models.

**Questions**
I'm uncertain about the implementation, which determines the orientation for my Tracab JSON deserializer. I'm confused again on HOME_TEAM / AWAY_TEAM / FIXED_HOME_AWAY.

In the meta data it is defined whether home or away team plays from left to right in the different periods. In the example file we have in our tests, we can read in the meta data that: `"Phase1HomeGKLeft": false` & `"Phase2HomeGKLeft": true`. This means that the home team GK has approx `"X": 5000` at the start of the first half and approx `"X": -5000` at the start of the second half, in the raw data.

What should the orientation then be? Should it be AWAY_TEAM for this example?

### DIFF
--- a/kloppy/_providers/tracab.py
+++ b/kloppy/_providers/tracab.py
@@ -1,8 +1,13 @@
-from typing import Optional
+from typing import Optional, Union, Type
+
 
 from kloppy.domain import TrackingDataset
-from kloppy.infra.serializers.tracking.tracab import (
-    TRACABDeserializer,
+from kloppy.infra.serializers.tracking.tracab.tracab_dat import (
+    TRACABDatDeserializer,
+    TRACABInputs,
+)
+from kloppy.infra.serializers.tracking.tracab.tracab_json import (
+    TRACABJSONDeserializer,
     TRACABInputs,
 )
 from kloppy.io import FileLike, open_as_file
@@ -15,8 +20,16 @@ def load(
     limit: Optional[int] = None,
     coordinates: Optional[str] = None,
     only_alive: Optional[bool] = True,
+    data_version: Optional[str] = None,
 ) -> TrackingDataset:
-    deserializer = TRACABDeserializer(
+    if data_version == "dat":
+        deserializer_class = TRACABDatDeserializer
+    elif data_version == "json":
+        deserializer_class = TRACABJSONDeserializer
+    else:
+        deserializer_class = identify_deserializer(meta_data, raw_data)
+
+    deserializer = deserializer_class(
         sample_rate=sample_rate,
         limit=limit,
         coordinate_system=coordinates,
@@ -28,3 +41,21 @@ def load(
         return deserializer.deserialize(
             inputs=TRACABInputs(meta_data=meta_data_fp, raw_data=raw_data_fp)
         )
+
+
+def identify_deserializer(
+    meta_data: FileLike,
+    raw_data: FileLike,
+) -> Union[Type[TRACABDatDeserializer], Type[TRACABJSONDeserializer]]:
+    deserializer = None
+    if "xml" in meta_data.name and "dat" in raw_data.name:
+        deserializer = TRACABDatDeserializer
+    if "json" in meta_data.name and "json" in raw_data.name:
+        deserializer = TRACABJSONDeserializer
+
+    if deserializer is None:
+        raise ValueError(
+            "Tracab data version could not be recognized, please specify"
+        )
+
+    return deserializer
--- a/kloppy/tests/files/tracab_raw.json
+++ b/kloppy/tests/files/tracab_raw.json
@@ -3686,6 +3686,478 @@
       ]
     },
     {
+      "FrameCount": 1942114,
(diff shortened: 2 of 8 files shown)


JanVanHaaren:
Based on @probberechts' table in #190, I presume the orientation in this particular example should be `FIXED_AWAY_HOME` or `AWAY_HOME` once #190 is merged.

probberechts:
> Based on @probberechts' table in #190, I presume the orientation in this particular example should be `FIXED_AWAY_HOME` or `AWAY_HOME` once #190 is merged.

That's right.

kloppy/tests/files/tracab_raw.json
```diff
@@ -0,0 +1,4406 @@
...
+        {
+          "BallContactInfo": "",
+          "BallOwningTeam": "H",
+          "BallStatus": "Dead",
+          "Speed": 121.28,
+          "X": 5550,
+          "Y": 4400,
+          "Z": 0
+        }
+      ]
```
DriesDeprest, line 2312: I don't really understand why we in the raw data, we have per frame 5 ball positions. It seems that the first ball position contains the right information, and the other 4 are duplicates but with the max X and Y values? Not sure what is going on.

DriesDeprest:
Any updates @JanVanHaaren @koenvo ?

kloppy/tests/files/tracab_raw.json
JanVanHaaren: Do your own data files have the same structure as this example data file?

kloppy/tests/test_tracab.py
```diff
@@ -18,19 +18,122 @@
...
+def xml_meta_data(base_dir: Path) -> Path:
+    return base_dir / "files" / "tracab_meta.xml"
+
+
+@pytest.fixture(scope="session")
+def dat_raw_data(base_dir: Path) -> Path:
+    return base_dir / "files" / "tracab_raw.dat"
+
+
+def test_correct_auto_recognize_deserialization(
```
JanVanHaaren, line 41: I think testing whether the `identify_deserializer` function returns the correct deserializer class would be better.

JanVanHaaren (commented):
Thank you for this pull request! While the approach generally looks good to me, I'm curious about the rationale behind supporting only two out of four possible scenarios in terms of meta data format and raw data format.

kloppy/tests/files/tracab_raw.json
DriesDeprest: Yes

kloppy/_providers/tracab.py
```diff
@@ -28,3 +41,21 @@ def load(
         return deserializer.deserialize(
             inputs=TRACABInputs(meta_data=meta_data_fp, raw_data=raw_data_fp)
         )
+
+
+def identify_deserializer(
```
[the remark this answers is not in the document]
DriesDeprest: My rationale was that I was going to use JSON meta and JSON raw data in my project and thus I added support for that. Didn't think someone would be interested in using XML meta and JSON raw data or JSON meta and DAT raw data.

kloppy/tests/files/tracab_raw.json
JanVanHaaren: Great. Thanks for checking with them!

kloppy/_providers/tracab.py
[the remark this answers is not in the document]
JanVanHaaren: Okay, that makes sense! I'm fine with keeping the current approach. If anyone else needs support for a different combination in the future, then the code could easily be refactored still.

JanVanHaaren (approved):
This pull request looks good to me now. Thanks!

koenvo:
Github tells me there is a merge conflict. Can you please make sure to merge the master in?

I really like the auto identification of the format to make it easier for users! Great work.

@JanVanHaaren I prefer `serialization_format` over `data_version`. What do you think? The version makes me think there can be different versions with different data in it, but it's only the format that changes, right?

DriesDeprest:
Thanks for the feedback @koenvo, should be good now!

probberechts:
I've noticed a small bug in this PR. Kloppy allows specifying the data source as a `FileLike` object but now it only works with a `Path`. For example with a string:

```py
from kloppy import tracab

dataset = tracab.load(
    meta_data="kloppy/tests/files/tracab_meta.xml",
    raw_data="kloppy/tests/files/tracab_raw.dat",
)
```

It gives the following error: `AttributeError: 'str' object has no attribute 'name'`

PR #297, opened 2024-02-24
Accurate transformation of pitch dimensions

### LINKED ISSUE #293: Different unit used for shot_event.result_coordinates.z dimension for Opta vs StatsBomb
**Opta**: The ground is 0 and the (middle of the) crossbar is at 40
**StatsBomb**: The ground is 0 and the (middle of the) crossbar is at 2.7

I assume StatsBomb expresses it in yards, for Opta I don't know.

How should we enable users to easily obtain standardized units for the result coordinates of a shot?

probberechts:
This PR is a complete rewrite of the "PitchDimensions" domain model. It implements the following main features.

1. It implements piecewise linear transformations in each dimension (see Stats Perform MA3 docs, appendix 7) to reduce the error when transforming between different coordinate systems.

```python
transformer = DatasetTransformer(
    from_pitch_dimensions=OptaPitchDimensions(),
    to_pitch_dimensions=MetricPitchDimensions(
        x_dim=Dimension(0, 105),
        y_dim=Dimension(0, 68),
        pitch_length=105,
        pitch_width=68,
        standardized=False,
    ),
)

transformed_point = transformer.change_point_dimensions(
    Point(17, 78.9)   # the corner of the penalty area...
)
# ... remains the corner of the penalty area
assert transformed_point == Point(16.5, 54.16)
```

2. It makes it possible to compute distances between two points without having to transform the entire dataset to a metric coordinate system.

```python
pitch = OptaPitchDimensions(pitch_length=105, pitch_width=68)
distance = pitch.distance_between(Point(0, 50), Point(11.5, 50))
assert distance == 11  # the distance from the goal line to penalty spot should be 11m
````

3. Fixes #293 
4. Allows creating pitch dimensions with unknown boundaries. This will allow solving #299 .

## BREAKING CHANGES: 
This changes the interface of the `PitchDimensions` class.

### DIFF
--- a/kloppy/domain/services/transformers/dataset.py
+++ b/kloppy/domain/services/transformers/dataset.py
@@ -239,23 +239,20 @@ class DatasetTransformer:
         if not point:
             return None
 
-        x = self._from_pitch_dimensions.x_dim.to_base(point.x)
-        y = self._from_pitch_dimensions.y_dim.to_base(point.y)
+        point_base = self._from_pitch_dimensions.to_base(point)
 
         if (
             self._from_coordinate_system.vertical_orientation
             != self._to_coordinate_system.vertical_orientation
         ):
-            y = 1 - y
+            point_base = replace(
+                point_base,
+                y=68 - point_base.y,
+            )
 
-        if not self._to_coordinate_system.normalized:
-            x = self._to_pitch_dimensions.x_dim.from_base(x)
-            y = self._to_pitch_dimensions.y_dim.from_base(y)
+        point_to = self._to_pitch_dimensions.from_base(point_base)
 
-        if isinstance(point, Point3D):
-            return Point3D(x=x, y=y, z=point.z)
-        else:
-            return Point(x=x, y=y)
+        return point_to
 
     def __flip_frame(self, frame: Frame):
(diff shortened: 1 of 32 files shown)

### CODE REVIEW

File: kloppy/domain/services/transformers/dataset.py [hunk 1]
```diff
@@ -245,23 +239,20 @@ def __change_point_coordinate_system(
...
+        point_base = self._from_pitch_dimensions.to_base(point)
 
         if (
             self._from_coordinate_system.vertical_orientation
             != self._to_coordinate_system.vertical_orientation
         ):
-            y = 1 - y
+            point_base = replace(
+                point_base,
+                y=68 - point_base.y,
```
koenvo, line 250: This seems like a bit of a magic number here. Can it be moved to some other place or to a 'constant' ?

koenvo:
One minor comment. Other than that happy to merge.

kloppy/domain/services/transformers/dataset.py
probberechts: I've replaced it by a constant + optimized it a bit more to avoid unneeded conversions between dimensions.

PR #298, opened 2024-02-25
Compute speed and acceleration

probberechts:
This PR adds support for computing the speed and acceleration of the ball and the players from tracking data.

### DIFF
--- a/kloppy/domain/models/common.py
+++ b/kloppy/domain/models/common.py
@@ -28,7 +28,17 @@ if sys.version_info >= (3, 11):
 else:
     from typing_extensions import Self
 
-from .pitch import PitchDimensions, Point, Dimension
+from .pitch import (
+    PitchDimensions,
+    Unit,
+    Point,
+    Dimension,
+    NormalizedPitchDimensions,
+    MetricPitchDimensions,
+    ImperialPitchDimensions,
+    OptaPitchDimensions,
+    WyscoutPitchDimensions,
+)
 from .formation import FormationType
 from ...exceptions import (
     OrientationError,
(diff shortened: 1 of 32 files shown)


ussf-jbekkers:
Great work @probberechts! 

Do you think there would be any merit to adding a separate smoothing window size for smoothing the ball trajectory, instead of using the same window size for player and ball?

probberechts:
@ussf-jbekkers I am not sure. Both the ball trajectory and player trajectories are measured using the same sensors, so I assume that they require the same filters. Yet, that assumption might be naive. Data providers might apply different post-processing to the ball and player trajectories. Also, I've never seen code or papers in which people applied different preprocessing to the ball and player trajectories. Do you have experience with applying different filters to the ball and player trajectories? If no one does this, I would not add 5 additional parameters.

UnravelSports:
I personally never considered this until I had a discussion with someone at Dagstuhl about this. His take was really simple, if you smooth the ball the same as you would player trajectories you will see a lot of those situations occur where the ball will "bend away" from a player when they make a pass (for example a one-touch pass) because the moment is pretty rapid and thus the sharp angle the ball would make in 4 or 5 frames will get smoothed so much that it looks like the ball floats away from the player. (I hope this makes sense verbally, I don't have visual example at hand). 
Additionally, even though the sensors are the same, the inherit movement of the ball is a lot more predictable and less prone to variation or "jitter" simply because the ball follows a (more or less) predetermined trajectory that doesn't change or deviate from a 2D/3D spline until it gets hit again. This is obviously very different behavior compared to human behavior that can be a lot more erratic. 

I understand not wanting to add it, especially because I also didn't consider it until quite recently and it seems like nobody has ever really cared, but it might at least be worth considering or preparing for by choosing parameter names that can be appended with additional/specific ball smoothing parameters without introducing breaking changes in the future.

probberechts:
@ussf-jbekkers Yes, that makes sense. I'll add the option to smooth the ball trajectory differently.

probberechts:
I came across this paper that seems interesting: https://journals.sagepub.com/doi/full/10.1177/17479541221124036

Issue #280, opened 2024-01-12
Creating support for expected goals and game state values in our kloppy data model


DriesDeprest:
**Expected goals**
By adding an optional `xg` attribute to our `ShotEvent` class, we can support the widely used expected goal property in kloppy. This property could be fed by the raw input data during deserialization (e.g. StatsBomb) or in a later stage could be calculated by the user using an xG model of choice.

Proposed implementation:

```python
@dataclass(repr=False)
@docstring_inherit_attributes(Event)
class ShotEvent(Event):
    """
    ShotEvent
    Attributes:
        event_type (EventType): `EventType.SHOT` (See [`EventType`][kloppy.domain.models.event.EventType])
        event_name (str): `"shot"`,
        result_coordinates (Point): See [`Point`][kloppy.domain.models.pitch.Point]
        result (ShotResult): See [`ShotResult`][kloppy.domain.models.event.ShotResult]
        xg (ExpectedGoal): See [`ExpectedGoal`][kloppy.domain.models.event.ExpectedGoal]
    """
    result: ShotResult
    result_coordinates: Point = None
    event_type: EventType = EventType.SHOT
    event_name: str = "shot"
    xg: Optional[ExpectedGoal]
 
    
@dataclass
class ExpectedGoal:
    """
    Expected goal metrics of an event

    Attributes:
         xg: The probability of scoring from the shot situation, not considering shot execution characteristics
         execution_xg: The probability of scoring following the execution of the shot
         gk_difficulty_xg: The probability of a goalkeeper conceding a goal
    """

    xg: Optional[float] = field(default=None)
    execution_xg: Optional[float] = field(default=None)
    gk_difficulty_xg: Optional[float] = field(default=None)

    @property
    def net_shot_execution(self) -> Optional[float]:
        return None if None in (self.xg, self.execution_xg) else self.execution_xg - self.xg
```

**Game state values**
By adding optional `gs_scoring_before`, `gs_scoring_after`, `gs_conceding_before` and `gs_conceding_after` attributes to our `Event` class, we can support the widely used game state based value models in kloppy. This property could be fed by the raw input data during deserialization (e.g. StatsBomb's on-the-ball value models) or in a later stage could be calculated by the user using a game state value model of choice (e.g. VAEP).

Proposed implementation:
```python
@dataclass
@docstring_inherit_attributes(DataRecord)
class Event(DataRecord, ABC):
    """
    Abstract event baseclass. All other event classes inherit from this class.

    Attributes:
[...]

PR #302, opened 2024-04-03
Add statistics attribute to DataRecord

DriesDeprest:
Related to: https://github.com/PySport/kloppy/issues/280.

I've started with adding support for xG statistics and wanted to get initial feedback on the implementation prior to expanding it to also parse OB values of StatsBomb and potentially other statistics.

So happy to hear your thoughts @probberechts @koenvo @JanVanHaaren


DriesDeprest:
It seems that for Python versions < 3.10, we get: TypeError: dataclass() got an unexpected keyword argument 'kw_only'.
But even for Python versions >= 3.10, we get: TypeError: non-default argument 'end_timestamp' follows default argument. But more generally this is the result of defining the default statistics argument in the `DataRecord` class prior to non-default arguments for classes which inherit from `DataRecord`, such as `Event`. 

I'm not sure what the right way is to handle this, so any help is greatly appreciated, thanks.

koenvo:
Check https://github.com/PySport/kloppy/blob/master/kloppy/domain/services/event_factory.py#L44-L49

DriesDeprest:
@koenvo @JanVanHaaren this implementation works for you guys?

kloppy/infra/serializers/event/statsbomb/specification.py
```diff
@@ -491,6 +493,15 @@ def _create_events(
...
             EVENT_TYPE.SHOT, shot_dict
         ) + _get_body_part_qualifiers(shot_dict)
+        shot_statistics = []
+
+        for statistic_cls, prop_name in {
+            ExpectedGoals: "statsbomb_xg",
+            PostShotExpectedGoals: "shot_execution_xg",
+        }.items():
+            value = shot_dict.get(prop_name, None)
+            if value:
```
koenvo, line 503: value can be zero. Guess is should be included when it’s not None

kloppy/domain/services/event_factory.py
```diff
@@ -66,7 +66,13 @@ def create_event(event_cls: Type[T], **kwargs) -> T:
             f"The following arguments were skipped: {skipped_kwargs}"
         )
 
-    return event_cls(**relevant_kwargs)
+    event = event_cls(**relevant_kwargs)
+
+    statistics = all_kwargs.get("statistics", [])
```
koenvo, line 71: I prefer this to be passed via kwargs, a couple of lines above. Like freeze_frame

DriesDeprest: I'm not sure how it would work.

```python
statistics: Optional[List[Statistic]] = field(
        init=False, default_factory=list
    )
```

In my implementation, the statistics cannot be initialized on creation, but only after. Thus, we cannot pass it as kwargs on creation, right?

The reason why I have put `init=False` is because otherwise it complains about `TypeError: non-default argument follows default argument`

DriesDeprest: A problem, however, with my current implementation is that the statistics of an event gets reset to [] when the dataclasses replace() method is being executed on an event as the result of e.g. a transformation. 

Any idea how to properly handle the statistics?

DriesDeprest: @koenvo could you review my latest code? Thanks!

kloppy/domain/services/transformers/dataset.py
```diff
@@ -196,57 +197,63 @@ def transform_frame(self, frame: Frame) -> Frame:
...
-                    ),
-                    distance=player_data.distance,
-                    speed=player_data.speed,
-                    other_data=player_data.other_data,
-                )
-                for key, player_data in frame.players_data.items()
-            },
-            other_data=frame.other_data,
+        frame = create_frame(
+            **dict(
```
koenvo, line 201: The `**dict` doesn't do anything. Can be:

```python
frame = create_frame(
    timestamp=frame.timestamp
    ...
)
```

kloppy/domain/services/transformers/dataset.py
```diff
@@ -196,57 +197,63 @@ def transform_frame(self, frame: Frame) -> Frame:
...
-                        player_data.coordinates
-                    ),
-                    distance=player_data.distance,
-                    speed=player_data.speed,
-                    other_data=player_data.other_data,
-                )
-                for key, player_data in frame.players_data.items()
-            },
-            other_data=frame.other_data,
+        frame = create_frame(
```
koenvo, line 200: Here the factory shouldn't be used. The `statistics` needs to be passed here to `Frame`

kloppy/domain/services/transformers/dataset.py
```diff
@@ -196,57 +197,63 @@ def transform_frame(self, frame: Frame) -> Frame:
...
+            **dict(
```
koenvo, line 231: The factory shouldn't be used here. The `statistics` needs to be passed here to `Frame`

kloppy/domain/services/transformers/dataset.py
```diff
@@ -292,18 +299,21 @@ def __flip_frame(self, frame: Frame):
...
-            timestamp=frame.timestamp,
-            frame_id=frame.frame_id,
-            ball_owning_team=frame.ball_owning_team,
-            ball_state=frame.ball_state,
-            period=frame.period,
-            # changes
-            ball_coordinates=self.flip_point(frame.ball_coordinates),
-            players_data=players_data,
-            other_data=frame.other_data,
+        frame = create_frame(
```
koenvo, line 302: The factory shouldn't be used here. The `statistics` needs to be passed here to `Frame`

kloppy/infra/serializers/event/statsbomb/helpers.py
```diff
@@ -146,15 +147,19 @@ def get_player_from_freeze_frame(player_data, team, i):
...
-            x=event.coordinates.x, y=event.coordinates.y, z=0
-        ),
-        players_data=players_data,
-        period=event.period,
-        timestamp=event.timestamp,
-        ball_state=event.ball_state,
-        ball_owning_team=event.ball_owning_team,
-        other_data={"visible_area": visible_area},
+    frame = create_frame(
+        **dict(
```
koenvo, line 151: `**dict(arg1=abc)` is the same as `arg1=abc`

DriesDeprest:
Resolved the comments, can you check again?

DriesDeprest:
@koenvo merge conflicts are resolved! Ready to merge!

kloppy/domain/models/common.py
```diff
@@ -888,6 +888,74 @@ class DatasetFlag(Flag):
...
+@dataclass
+class PostShotExpectedGoals(ScalarStatistic):
+    """Post-shot expected goals"""
+
+    def __post_init__(self):
+        self.name = "PSxG"
+
+
+@dataclass
+class GameStateValue(Statistic):
```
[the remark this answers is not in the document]
JanVanHaaren: I agree that `GameStateValue` might not be the most appropriate name. Assuming that instances of this class are attached to individual events or actions, I think `ActionValue` would be more suitable than `PossessionValue`.

kloppy/domain/models/common.py
```diff
@@ -888,6 +888,74 @@ class DatasetFlag(Flag):
...
+    """Game state value"""
+
+    gsv_scoring_before: Optional[float] = field(default=None)
+    gsv_scoring_after: Optional[float] = field(default=None)
+    gsv_conceding_before: Optional[float] = field(default=None)
+    gsv_conceding_after: Optional[float] = field(default=None)
+
+    def __post_init__(self):
+        self.name = "GSV"
```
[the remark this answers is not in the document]
probberechts: Would it be possible to set the name used by the dataprovider here? Like "OBV" for StatsBomb, "PV" for StatsPerform, etc.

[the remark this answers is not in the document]
DriesDeprest: Does my implementation make sense for setting the name based on providers?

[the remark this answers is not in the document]
probberechts: Isn't it possible to do something like: `action_value = ActionValue(name="OBV", **game_state_values_data)` in the StatsBomb parser? I think it should be possible to attach multiple action values to a dataset. For example, one might also want to compute VAEP for StatsBomb data.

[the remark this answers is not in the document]
DriesDeprest: It indeed makes more sense to just directly pass the name to the ActionValue object. I've adjusted the code accordingly.

Issue #277, opened 2024-01-08
Store the position of a player at moment of event execution and not only starting position


DriesDeprest:
**Current**
Currently, we only store the starting position of a player in our `EventDataset.metadata.teams[x].players[x].position`. It is also this starting position of the player that we add to every event `EventDataset.events[x].player.position`. 

**Goal**
I believe it would be more accurate to store the position of the player at the time of execution in every event `EventDataset.events[x].player.position`, instead of the starting position.

**Implementation**
For some providers (e.g. Wyscout & StatsBomb) we can derive the position of the player at the moment of the event directly from the raw data.
While for others (e.g. Opta) we would have to derive the position of the players from the starting positions and formation change events in the serializer. In this case, we would update the positions of the different players throughout serializing the raw file so we can use the correct position for each player throughout serialization.

JanVanHaaren:
It would indeed be valuable to keep track of the actual positions of the players throughout a match. In the StatsBomb deserializer, we currently use the _Tactical Shift_ events only to update the teams' formations but we could also use them to update the players' positions.

DriesDeprest:
I was originally planning to not make use of the _Tactical Shift_ events for StatsBomb, but just use the "position" property of each event in the raw data, to set the right position for the player for every event.

This would look something like this:
``` python
player = team.get_player_by_id(raw_event["player"]["id"] if "player" in raw_event else None
if player:
  player.position = Position(position_id=raw_event["position"]["id"], name=raw_event["position"]["name"])
generic_event_kwargs = {..., "player": player, ...}
```
For Wyscout, a similar approach could be used. To make this work, however, we would need to remove the `(frozen=True)` argument from the `@dataclass` decorator wrapping the `Player` class ([context](https://github.com/PySport/kloppy/blob/master/kloppy/domain/models/common.py#L116-L142)). Do you see any problems with that? 

Only for Opta, we would have to derive the position of the players from the starting positions and formation change events in the serializer, as the position of the player is not present in the raw data.

probberechts:
@JanVanHaaren 
>  In the StatsBomb deserializer, we currently use the Tactical Shift events only to update the teams' formations

To be precise, we do not update the `Team`. Instead, we create a `FormationChangeEvent` with a `formation_type` field. From this, we can derive the team's formation in each event using the `FormationStateBuilder`.

@DriesDeprest 
[...]

Issue #301, opened 2024-03-27
Store the formation changes of a team in Team object


DriesDeprest:
I would like to create a `formation_changes` attribute in our `Team` class. I would use it to make it easier to calculate how much a team has played in a given formation in downstream usage of the `EventsDataset`.

My proposal would be to make a similar change as we are planning to do for player positions, where we are moving from having a (starting) `position` attribute to a `position_changes` attribute, with a `starting_position` and `position` method to easily retrieve the position at any given moment. The details of these changes are discussed in: https://github.com/PySport/kloppy/issues/277.

So the changes to our Team class, would look something like this:
```python
@dataclass(frozen=True)
class FormationChange:
    timestamp: timedelta
    period_id: int
    formation: Formation

@dataclass
class Team:
    team_id: str
    name: str
    ground: Ground
    formation_changes: List[FormationChange] = field(default_factory=list)
    players: List[Player] = field(default_factory=list)
    
    @property
    def starting_formation(self):
        if self.formation_changes:
            return self.formation_changes[0].formation

    def formation(self, period_id: int, timestamp: timedelta):
        if self.formation_changes:
            formation = self.formation_changes[0].formation
            for formation_change in self.formation_changes:
                # Check if the current change is before or exactly at the given period and timestamp
                if period_id > formation_change.period_id or (
                    period_id == formation_change.period_id
                    and timestamp >= formation_change.timestamp
                ):
                    formation = formation_change.position
                else:
                    # Since the list is assumed to be in chronological order, break on finding the first future change
                    break

            return formation
```

However, I am bit struggling with how to think about the overlap of functionality between the `add_state` method with 'formation' argument and the new `formation` method.  They seem to be both serving a similar functionality, determining the formation of team at a given timepoint in the game. Maybe, their purposes are slightly different and it is deseriable to have the both of them in kloppy?

@probberechts @JanVanHaaren @koenvo thoughts?

koenvo:
Fixed in https://github.com/PySport/kloppy/pull/332

PR #307, opened 2024-04-08
Track player position changes

DriesDeprest:
Created this draft PR related to: https://github.com/PySport/kloppy/issues/277, to get initial feedback on the implementation prior to starting updating all the deserializers. This is also based on the discussion made here: https://github.com/probberechts/kloppy/pull/1. 

As a sake of example, I've started with updating the Statsomb deserializer to support the Player.positions.

@JanVanHaaren @koenvo happy to hear what you guys think.

### DIFF
--- a/kloppy/domain/models/common.py
+++ b/kloppy/domain/models/common.py
@@ -126,6 +126,13 @@ class Position:
 
 
 @dataclass(frozen=True)
+class PositionChange:
+    timestamp: timedelta
+    period_id: int
+    position: Position
+
+
+@dataclass(frozen=True)
 class Player:
     """
     Attributes:
@@ -135,8 +142,7 @@ class Player:
         name: Full name of the player
         first_name: First name
         last_name: Last name
-        starting: `True` when player is part of the starting 11
-        position: See [`Position][kloppy.domain.models.common.Position]
+        position_changes: See [`PositionChange][kloppy.domain.models.common.PositionChange]
         attributes: attributes given by the provider
     """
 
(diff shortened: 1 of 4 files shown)


DriesDeprest:
@koenvo @JanVanHaaren did you guys find the time to take a look this PR? I'd like to continue the effort, but first want to check whether you agree with the implementation proposal.

JanVanHaaren:
Thank you for your proposal, @DriesDeprest! Your approach generally looks good to me.

The following thoughts crossed my mind while looking through the code.

1. I see similarities with the formation changes for teams. Would the position changes be used in similar way?
2. Would it be useful to know the end timestamp for each position in addition to the start timestamp?
3. Are the timestamps ever in the `mm:ss` format instead of the `HH:mm:ss` format?

What are your thoughts, @probberechts?

DriesDeprest:
Thanks for reviewing @JanVanHaaren!

1.

I noticed the same similarities and raised it here: https://github.com/PySport/kloppy/issues/301. How I see it, you have two options on how to get the actual formation / position of a team / player for each event.
a) You create a specific event (FormationChange/PositionChange) for it and using the state builder, you can get the state formation/position at time of execution of the event.
b) You have a formation/position_change attribute in the Team/Player object, instead of a formation/position attribute, and with the right method position/formation() you can get the actual formation / position of a team / player for each event.

For formations, we have historically chosen to go for implementation a) but for positions we are now building implementation b). 

The reason why I prefer option b) over option a) is because it also offers you a standardized data format for storing formation changes of a team and position changes of players of a football game across vendors, besides the ability to get the actual formation / position of a team / player for each event 

2. See https://github.com/PySport/kloppy/issues/277#issuecomment-2015474935 for why we don't store the end timestamp of a position change.

3. I believe so, but will double-check once I continue the implementation, after we agree on the high level implementation.

DriesDeprest:
Closing this PR, since we added position changes with: https://github.com/PySport/kloppy/pull/326.

PR #308, opened 2024-04-12
[IO] Improved IO with support for reading data from compressed files

probberechts:
It is a common practice to store data as compressed files to reduce storage requirements. With this PR it is no longer needed to decompress the file before loading the data with kloppy.

```py
from kloppy import statsperform

dataset = statsperform.load(
    raw_data="ma25_tracking.txt.gz",
    meta_data="ma1_metadata.xml.gz",
)
```

Whether a file is compressed is derived from the file's extension. Currently supports ".gz", ".xz" and ".bz2".


kloppy/io.py
```diff
@@ -35,10 +47,12 @@ def create(cls, input_: "FileLike", **kwargs):
 
 
 def get_file_extension(f: FileLike) -> str:
-    if isinstance(f, str):
+    if isinstance(f, PurePath) or isinstance(f, str):
+        f = str(f)
+        for ext in [".gz", ".xz", ".bz2"]:
```
koenvo, line 52: Nice addition! Would be good to have a short description of this behaviour.

koenvo:
This should also work non-local files, right? Like `https://some-url.com/file.xml.gz`

koenvo:
Can you merge master in please to make sure tests run again

probberechts:
I couldn't get boto (to mock an S3 bucket) to work on GitHub Actions. In the most recent version, there is [this bug](https://github.com/aio-libs/aiobotocore/issues/755) and for older versions I can't figure out a set of version constraints between s3fs and boto that works on each Python version. Hence, I propose to disable these tests until the bug is fixed.

I recently also found the [xopen](https://github.com/pycompression/xopen) library for opening compressed files. We could use it as a more efficient and robust replacement of the `_open` method that I implemented. Do you think it is worth adding another dependency? It could also be an optional one.

kloppy/io.py
```diff
@@ -1,25 +1,79 @@
...
+    return open(file, mode)
+
+
+def _decompress(stream: BinaryIO) -> BinaryIO:
+    stream.seek(0)
+    if stream.read(3) == b"\x1f\x8b\x08":
+        import gzip
+
+        stream.seek(0)
+        print("GZIP")
```
koenvo, line 48: Can you remove the print statement please

kloppy/io.py
```diff
@@ -1,25 +1,79 @@
...
+    elif file.endswith(".bz2"):
+        import bz2
+
+        return bz2.open(file, mode)
+    return open(file, mode)
+
+
+def _decompress(stream: BinaryIO) -> BinaryIO:
+    stream.seek(0)
+    if stream.read(3) == b"\x1f\x8b\x08":
```
koenvo, line 44: Haven't tested this yet, but this should not work, right?

The first `read(3)` will change the file pointer, which will cause the next `read(6)` to start at offset 3.

probberechts: Indeed, good catch! It has been fixed now.

koenvo:
Thanks Pieter, great work!

PR #309, opened 2024-04-12
[StatsPerform] Bugfixes for tracking deserializer

### LINKED ISSUE #299: Incorrect Stats Perform coordinate system
Kloppy defines the following coordinate system for Stats Perform tracking data:

```
origin = BOTTOM_LEFT
vertical_orientation = BOTTOM_TO_TOP
x_dim = [0,100]
y_dim = [0,100]
```

However, this does not seem to correspond with any of the 4 coordinate systems used by Stats Perform.

image

Also, it is currently not possible to implement any of these coordinate systems in the data model of kloppy, because the true length and width of the pitch are not defined in the source data.

probberechts:
- Use the correct coordinate system and pitch dimensions (fixes #299).
- Remove the "BALL_OWNING_TEAM" flag. The Stats Perform tracking data does not define which team has possession.
- The `periods` field of the metadata should be a list, not a dict.
- Skip frames in which the game is paused when `only_alive = True`.


probberechts:
Superseded by #310.

Issue #312, opened 2024-05-14
Periods creation for StatsBomb deserializer fails in rare occasions


DriesDeprest:
In rare occasions, I've seen that StatsBomb matches have more than one period start / end event for a given team in their events data, which results in the StatsBomb deserializer returning a `DeserializationError: "Failed to determine start and end time of periods."`.

When this is the case the `first_start_and_end_half_events` contains 5 elements (see screenshot below), which results into the `DeserializationError`.

![image](https://github.com/PySport/kloppy/assets/46134788/765cb47b-ad12-41cd-91a0-b38b80ced6d6)

PR #313, opened 2024-05-14
Fix StatsBomb create_periods

DriesDeprest:
Wrote a fix for: https://github.com/PySport/kloppy/issues/312

Updated the create periods method of StatsBomb parser, so it does not fail if in the raw event data multiple period start or period end events are present for the same period and team.


kloppy/infra/serializers/event/statsbomb/deserializer.py
```diff
@@ -216,17 +216,39 @@ def create_team(lineup, ground_type):
         return [home_team, away_team]
 
     def create_periods(self, raw_events):
```
JanVanHaaren, line 218: I think the implementation of this method could be simplified.

**Example**
```python
    def create_periods(self, raw_events):
        half_start_events = {}
        half_end_events = {}
        for event in raw_events.values():
            event_type = SB.EVENT_TYPE(event.raw_event["type"])
            period = event.raw_event["period"]

            if event_type == SB.EVENT_TYPE.HALF_START:
                half_start_events[period] = event.raw_event
            elif event_type == SB.EVENT_TYPE.HALF_END:
                half_end_events[period] = event.raw_event

        periods = []
        for start_event, end_event in zip_longest(
            half_start_events.values(), half_end_events.values()
        ):
            if start_event is None or end_event is None:
                raise DeserializationError(
                    "Failed to determine start and end time of periods."
                )
            start_timestamp = (
                periods[-1].end_timestamp
                + parse_str_ts(start_event["timestamp"])
                if len(periods) > 0
                else parse_str_ts(start_event["timestamp"])
            )
            end_timestamp = (
                start_timestamp + parse_str_ts(end_event["timestamp"])
                if end_event
                else None
            )
            period = Period(
                id=int(start_event["period"]),
                start_timestamp=start_timestamp,
                end_timestamp=end_timestamp,
            )
            periods.append(period)
        return periods
```

JanVanHaaren (changes requested):
Thank you for the fix! I suggested a simpler implementation.

kloppy/infra/serializers/event/statsbomb/deserializer.py
DriesDeprest: Thanks, looks indeed more elegant. Made the requested changes.

PR #321, opened 2024-05-22
Missing "Result" Attribute makes Live Opta parsing impossible

UnravelSports:
Hi! 

A really simple fix here. When loading opta event data (as shown below) for an incomplete f7 (read: during live matches) we get an AttributeError. 
```python
dataset = opta.load(
    f7_data="f7.xml",
    f24_data="f24.xml"
)
``` 

Error looks as follows:
```shell
Traceback (most recent call last):
  File "/workspaces/kloppy/_dev.py", line 3, in <module>
    dataset = opta.load(
  File "/workspaces/kloppy/kloppy/_providers/opta.py", line 35, in load
  File "/workspaces/kloppy/kloppy/infra/serializers/event/opta/deserializer.py", line 729, in deserialize
    match_result_type = list(match_result_path.find(f7_root))[
  File "src/lxml/objectpath.pxi", line 55, in lxml.objectify.ObjectPath.__call__
  File "src/lxml/objectpath.pxi", line 219, in lxml.objectify._find_object_path
AttributeError: no such child: Result
``` 

Normally, in a completed game we'd have a Result tag that looks something like this:
```html
<Result Type="Aggregate" Winner="t12345" />
``` 

But, because the game is not finished yet there simply is no "Result" tag.

Doing a simple try / except block like below resolves this issue and creates a correct Kloppy EventDataset.
```python
try:
              match_result_type = list(match_result_path.find(f7_root))[
                  0
              ].attrib["Type"]
          except AttributeError:
              match_result_type = None
``` 

Not 100% sure if using a try-except block here is the best solution, but it works.


UnravelSports:
Great, it's still including some stupid commits from a year ago that are totally irrelevent.....

PR #323, opened 2024-05-28
[Wyscout v3] add parsing of carry event 

DriesDeprest:
Add support for the creation of a carry event for Wyscout v3 event data.

"carry" is a secondary event type, thus I've added the potential recognition of a "carry" event at the bottom of the if / elif statements for recognizing the event type. This way, we give priority to primary event types for determining the kloppy event type.

We derive the end timestamp as the timestamp of the next event, if it is present, otherwise it is said equal to the timestamp.


JanVanHaaren (approved):
Thank you! This pull request looks good to me.

PR #326, opened 2024-06-11
Feature/initial player position

koenvo:


kloppy/domain/models/common.py
```diff
@@ -126,6 +126,10 @@ class Position:
     def __str__(self):
         return self.name
 
+    @classmethod
+    def unknown(cls) -> "Position":
+        return cls(position_id="", name="Unknown")
+
 
 @dataclass(frozen=True)
 class Player:
```
probberechts, line 135: This class now has `initial_position`, `starting_position` and `position` attributes. I don't see immediately what's the difference between `initial_position` and `starting_position`. The implementation of the `position` attribute is also counterintuitive to me. I would expect that it returns a player's initial position, not the last position. 

I think you should be able to easily retrieve a players position at the start of the game and the list of subsequent position changes. The last position isn't more important than any other position. Hence, my suggestion would be to keep `starting_position` and make `position` identical to `starting_position` but deprecate it. I prefer `starting_position` over `initial_position`.

kloppy/domain/models/event.py
```diff
@@ -1060,6 +1062,34 @@ class EventDataset(Dataset[Event]):
 
     dataset_type: DatasetType = DatasetType.EVENT
 
+    def _update_player_positions(self):
+        """Update player positions based on"""
```
probberechts, line 1066: Based on what? :smiley:

kloppy/domain/models/time.py
```diff
@@ -0,0 +1,262 @@
...
+        return f"P{self.period.id}T{m:02.0f}:{s:02.0f}"
+
+    def __hash__(self):
+        return hash((self.period.id, self.timestamp.total_seconds()))
+
+
+T = TypeVar("T")
+
+
+class TimeContainer(Generic[T]):
```
probberechts, line 202: It would be handy to have a `__repr__` method such that you can easily print a player's positions throughout a game.

probberechts (commented):
Nice contribution @koenvo! I only have a few minor suggestions.

kloppy/domain/models/common.py
koenvo: Agree on the confusion naming. The `initial_position` is intended as the "position from the metadata". This can also be a position from a non-starting player.

I'll update and follow your idea. Thanks!

kloppy/domain/models/event.py
koenvo: Hah! Good catch

kloppy/domain/models/time.py
koenvo: Yes!

JanVanHaaren (approved):
Great change! This pull request looks good to me.

probberechts (approved):

Issue #315, opened 2024-05-21
Wyscout v3 events sometimes have 10 player formations for (opponent) team


DriesDeprest:
I noticed that the team / opponent team object in the events of Wyscout v3 event data, doesn't always have a formation where the total amount of players summed up equals to 11. For example, I've seen occurrences of events with a "4-4-1" (opponent) team formation. Currently, our serializer crashes when this is the case, as it does not recognize this formation.

When analysing the event data of a match where we have the 'troublesome' formations, I saw that this was the result of a team getting a red card booking and in the events that followed the formation of that team was described by a 10 player formation.
The team originally had a "4-4-1-1" formation, but after the red card this shifted to "4-4-1".

How do we want to handle this?

**Option A:**
- Description: We create new generic kloppy formation types for all possible 10 player formations present in Wyscout and use those to describe the formation of the (opponent) team in the data. 
- Pro: This keeps the kloppy output close to the raw data input.
- Con: The behaviour of how we describe 10 player formations is different for Wyscout vs other providers. Where for other providers, we still use a 11 player formation after a red card. Resulting in a non-standardized approach for different providers.

**Option B:**
- Description: When we recognize a 10 player formation in the event data, we keep using the last valid 11 player formation observed in the data of that team for all future events of that team.
- Pro: This keeps the behaviour standard over different providers, where after a red card we still use 11 player formations to describe team formations.
- Con: We lose a level of detail of describing the formations of a team.

I think my preference would go to option B, to have standard behaviour across different providers.

dvilches:
In my personal opinion, it is always better for the data to reflect reality as accurately as possible.

In our case, it is important to know our own and our opponent's formation, as we analyze "behavior" with different schemes and clearly when there is one or more players less on the field, that changes.

Likewise, it would have to be seen what most users use Kloppy for, since these "own" issues can be solved, as until now, by performing our own processing on eventing data, in this case.

DriesDeprest:
@koenvo @JanVanHaaren @probberechts  thoughts?

DriesDeprest:
@dvilches thanks for sharing your take on option A vs B. I understand your need of having an accurate description of a team's behaviour to perform qualitative performance analysis.

Since I'm using kloppy for reading in data from different providers, the aspect that we have a standardized output for different input vendors is more importantly for my use case than the level of detail that we get extra. Therefore, my preference for option B.

[...]

PR #330, opened 2024-06-25
[Wyscout v3] Set formation to `FormationType.UNKNOWN` when a 10 player formation is recognized

DriesDeprest:
See https://github.com/PySport/kloppy/issues/315. This is an implementation of option B explained in this issue.


DriesDeprest:
@probberechts any thoughts on this? :-)

probberechts:
@DriesDeprest Can you clean up the old code? E.g., remove all references to `last_formations`

DriesDeprest:
@probberechts should be good now!

PR #332, opened 2024-07-03
Add formations attribute to Team class

DriesDeprest:
I would like to add the formation changes of a team during a game in the `Dataset` of that game. Thus, I've added a `formations` attribute to the `Team` class, to contain this information.

The implementation is similar to that of the `positions` attribute in the `Player` class.


koenvo:
Looks good to me!

koenvo (approved):

PR #338, opened 2024-07-15
Fix/skillcorner timestamp

UnravelSports:
I noticed that at some point SkillCorner changed all their "time" keys to "timestamp", not only in the "jsonl" files.

So, I rewrote `__load_json()` and added `__load_json_raw()` the latter renames any "timestamp" key back to "time" to make it compatible with every time the "time" key is required in the deserializer.

Also added a test and a test file for the "timestamp" situation. This test simply loads the dataset, the rest should be the same as the other tests.

Note: I removed a lot of 2nd half frames from the test file, because the file size was greater than 50MB.

### DIFF
--- a/kloppy/infra/serializers/tracking/skillcorner.py
+++ b/kloppy/infra/serializers/tracking/skillcorner.py
@@ -220,13 +220,13 @@ class SkillCornerDeserializer(TrackingDataDeserializer[SkillCornerInputs]):
                 data.append(obj)
             return data
         else:
-            data = json.load(file)
+            data = self.__load_json(file)
             for line in data:
                 if "timestamp" in line:
                     line["time"] = line.pop("timestamp")
             return data
-        
-    def __load_json_meta(self, file):
+
+    def __load_json(self, file):
         return json.load(file)
 
     @classmethod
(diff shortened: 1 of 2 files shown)


kloppy/infra/serializers/tracking/skillcorner.py
```diff
@@ -209,7 +209,7 @@ def _get_skillcorner_attacking_directions(cls, frames, periods):
 
         return attacking_directions
 
-    def __load_json(self, file):
+    def __load_json_raw(self, file):
```
JanVanHaaren, line 212: The refactored function looks a bit messy to me.

1. Should we avoid code duplication by moving the timestamp fix into an auxiliary function?
2. Do we need the `__load_json` auxiliary function? In the `if` branch, the `json.loads` function is called directly.

UnravelSports: I agree that it was a bit messy, latest commit should've addressed all your concerns.

- removed `__load_json`, not sure why it was there to begin with
- added `__replace_timestamp`

JanVanHaaren (approved):
This pull request looks good to me! Thank you.

PR #349, opened 2024-08-15
[StatsPerform] Support frames with no player data

DriesDeprest:


koenvo:
Thanks for this PR. I'm wondering if it should result in an empty Frame, or if the Frame should be left out. In what cases do you get data without player data?

DriesDeprest:
I don't know what resulted in the raw tracking data not containing any positional information of the players. In the tracking data file where I observed this, it was the last 20 or rows / lines that only contained positional information of the ball.

Including an empty Frame or excluding the Frame with no positional data of the players, both work for me.

Since it seems to be rare for the tracking data of a match to contain lines without positional data of the player, I've only observed it in one file of a set of 50, and it to only affect a few lines at the end of the match, I don't think it matters that much. Most important is that the deserializer just doesn't break, I'd say.

UnravelSports:
I know it seems to be a rare occurrence, and because of that the StatsPerform serializer doesn't take a `include_empty_frames` parameter like for example SkillCorner does. It does feel like we need to do something with this though, because even though it doesn't happen very often, simply loading in partial frames can break other stuff down the line. Not sure what the correct approach is though.

koenvo:
Would stuff break when we just leave those frames out? And maybe later on add a Data Quality Report (or something) that includes information about missing/incomplete data that has been filtered out

koenvo:
Can you update the PR so it will leave empty frames out by default?

DriesDeprest:
done

PR #353, opened 2024-09-27
Tracab Meta added (V4)

UnravelSports:
I've added an option for reading yet another version of the Tracab Meta data (as requested by here https://github.com/PySport/kloppy/discussions/352)

My main concern is the ever growing amount of random Tracab Meta file configurations, but for now this implementation should hold. Not sure how to approach this other than wait for issues to be raised...

It first checks if there is a "match" attribute in the meta_data and if not it will check a key I have yet only seen in this new version of the meta data (`Phase1StartFrame`). Additionally there are some small additions to make the create_team function a bit more flexible because some tags have slightly changed in this version. `<Player>` became `<item>` and `<TeamId>` became `<TeamID>` (same for player ids).

I've also added new tests.


UnravelSports:
The tests seem to fail at a random file I didn't touch?

Alek050:
> The tests seem to fail at a random file I didn't touch?

Was just sniffing around this issue. Seems like the git actions fail on the formatting of `kloppy/infra/serializers/event/sportec/deserializer.py`. I am guessing that your local formatting settings (maybe with black?) are different from the ones on git, but not sure though.

UnravelSports:
I see my commit contains that file, in a slightly different format. Weird 🤷‍♀️

UnravelSports:
I had black version 23.3.0 instead of 22.3.0...

PR #355, opened 2024-10-01
[StatsPerform - Tracking] Handle missing end period for abandoned matches

DriesDeprest:
When a match is abandoned, the "end" attribute misses in the raw data. We handle that more gracefully and in case the datetime is present of when the match is abandonded we use that for the end of the period.


JanVanHaaren:
The code change looks good to me. Could you please add a test for this functionality?

DriesDeprest:
done

PR #356, opened 2024-10-23
[Signality] adding support for Signality tracking data

DriesDeprest:


DriesDeprest:
@koenvo I don't know why the last CI / CD run failed here, can you trigger a retry?

kloppy/_providers/signality.py
```diff
@@ -0,0 +1,39 @@
+from typing import Optional, Union
```
probberechts, line 1:
```suggestion
from typing import Optional
```

kloppy/_providers/signality.py
```diff
@@ -0,0 +1,39 @@
...
+    SignalityDeserializer,
+    SignalityInputs,
+)
+from kloppy.io import FileLike, open_as_file
+
+
+def load(
+    meta_data: FileLike,
+    p1_raw_data: FileLike,
+    p2_raw_data: FileLike,
```
probberechts, line 14: This will only work for matches with two phases (i.e., periods). I'm not familiar with Signality data but I assume that a game can have more than two phases if it goes to overtime. I think we should deal with this in the same way as we do for HawkEye data.

kloppy/_providers/signality.py
```diff
@@ -0,0 +1,39 @@
...
+
+from kloppy.domain import TrackingDataset
+from kloppy.infra.serializers.tracking.signality import (
+    SignalityDeserializer,
+    SignalityInputs,
+)
+from kloppy.io import FileLike, open_as_file
+
+
+def load(
```
probberechts, line 11: Can you add a docstring for this method? To me it is not obvious what the inputs (in particular `meta_data` and `venue_information`) are.

kloppy/infra/serializers/tracking/signality.py
```diff
@@ -0,0 +1,267 @@
+import logging
+from datetime import timedelta, datetime
+import warnings
+from typing import List, Dict, NamedTuple, IO, Optional, Union
+from collections import Counter
+import numpy as np
```
probberechts, line 6: We don't need / want numpy as a dependency.

kloppy/infra/serializers/tracking/signality.py
```diff
@@ -0,0 +1,267 @@
...
+            period_id = period.id
+            if period_id in frame_period_ids:
+                count = Counter(
+                    frame_attacking_directions[frame_period_ids == period_id]
+                )
+                attacking_directions[period_id] = count.most_common()[0][0]
+            else:
+                attacking_directions[period_id] = AttackingDirection.NOT_SET
+
+        return attacking_directions
```
probberechts, line 139: This code seems to be copy-pasted from the SkillCorner deserializer. Can you move this to kloppy/domain/services/__init__.py?

probberechts:
Nice contribution! I am not familiar with Signality data so it would be good if someone else could take a look at this as well, but I've already left a few suggestions for improving the code.

kloppy/_providers/signality.py
DriesDeprest: I've reworked it, so raw tracking data of a match are passed as a list of raw tracking data per period. This way any arbitrary number of periods are supported.

kloppy/infra/serializers/tracking/signality.py
DriesDeprest: We use numpy in the `attacking_directions_from_multi_frames()` function which I now moved to kloppy/domain/services/init.py. Is that okay? I assumed so since we were already using it in our skillcorner implementation.

probberechts: I fixed that yesterday in #375

DriesDeprest:
I've resolved merge conflicts. Anything else I can help with to get this merged @probberechts?

DriesDeprest:
I don't know why these steps in the CI/CD process failed

image

@probberechts do you maybe know why?

probberechts:
> I don't know why these steps in the CI/CD process failed
> @probberechts do you maybe know why?

It's an issue with one of the dependencies of the docs. See https://github.com/lepture/mistune/issues/403 I think we should be able to fix this by pinning the version in docs-requirements.txt or simply wait a bit until it gets fixed:

```
mistune==3.0.2 # temporary pin until https://github.com/lepture/mistune/issues/403 is resolved.
```

probberechts:
I was waiting for @UnravelSports to complete #408 as HawkEye also requires support for providing a list of data files as input. So, let me first check that one and then we should make sure that we handle it consistently for both HawkEye and Signality.

probberechts:
@DriesDeprest Could you update the code to use the `io.expand_inputs` function? You can look at #408 for an example.

kloppy/_providers/signality.py
```diff
@@ -0,0 +1,56 @@
...
+        sample_rate (Optional[float]): Sampling rate to be applied during deserialization (default: None).
+        limit (Optional[int]): Limit on the number of data points to process (default: None).
+        coordinates (Optional[str]): Coordinate system to use for deserialization (default: None).
+
+    Returns:
+        TrackingDataset: A deserialized tracking dataset object.
+
+    This function opens the provided metadata, raw data feeds, and venue information files, and
+    uses the SignalityDeserializer to process them into a TrackingDataset object. All file
+    resources are properly managed to ensure they are closed after use.
```
probberechts, line 36: This seems useless information.

kloppy/_providers/signality.py
```diff
@@ -0,0 +1,56 @@
...
+    venue_information: FileLike,
+    sample_rate: Optional[float] = None,
+    limit: Optional[int] = None,
+    coordinates: Optional[str] = None,
+) -> TrackingDataset:
+    """
+    Load and deserialize tracking data from multiple input files.
+
+    Args:
+        meta_data (FileLike): File-like object containing metadata information of the game.
```
probberechts, line 24: "A file-like object" is very vague. It is better to refer to something that the user is familiar with. For example, here I would write "json feed containing the metadata information of the game"

kloppy/_providers/signality.py
```diff
@@ -0,0 +1,56 @@
...
+from kloppy.io import FileLike, open_as_file, expand_inputs
+
+
+def load(
+    meta_data: FileLike,
+    raw_data_feeds: Iterable[FileLike],
+    venue_information: FileLike,
+    sample_rate: Optional[float] = None,
+    limit: Optional[int] = None,
+    coordinates: Optional[str] = None,
```
probberechts, line 18: I believe the `coordinates` parameter accepts both a string and a `CoordinateSystem` object. This seems to be wrong for a lot of providers, though.

kloppy/_providers/signality.py
```diff
@@ -0,0 +1,56 @@
...
+) -> TrackingDataset:
+    """
+    Load and deserialize tracking data from multiple input files.
+
+    Args:
+        meta_data (FileLike): File-like object containing metadata information of the game.
+        raw_data_feeds (Iterable[FileLike]): An iterable of file-like objects containing raw tracking data feeds.
+        venue_information (FileLike): File-like object containing venue information where the game was played.
+        sample_rate (Optional[float]): Sampling rate to be applied during deserialization (default: None).
+        limit (Optional[int]): Limit on the number of data points to process (default: None).
```
probberechts, line 28: What is a "data point" here: a single player's position or a frame? Better write "limit the number of frames loaded"

kloppy/domain/models/common.py
```diff
@@ -836,6 +837,44 @@ def pitch_dimensions(self) -> PitchDimensions:
             )
 
 
+@dataclass
+class SignalityCoordinateSystem(CoordinateSystem):
```
probberechts, line 841: Has to be updated to 
```
class SignalityCoordinateSystem(ProviderCoordinateSystem):
```

kloppy/infra/serializers/tracking/signality.py
```diff
@@ -0,0 +1,241 @@
+import logging
+from datetime import timedelta, datetime
+import warnings
+from typing import List, Dict, NamedTuple, IO, Optional, Union, Iterable
+from collections import Counter
```
probberechts, line 5: Unused import

kloppy/infra/serializers/tracking/signality.py
```diff
@@ -0,0 +1,241 @@
...
+import numpy as np
+import json
+
+from kloppy.domain import (
+    attacking_direction_from_frame,
```
probberechts, line 10: Unused import

kloppy/infra/serializers/tracking/signality.py
```diff
@@ -0,0 +1,241 @@
...
+    AttackingDirection,
+    Frame,
```
probberechts, line 12: Unused import

kloppy/infra/serializers/tracking/signality.py
probberechts: But  numpy is still imported here

kloppy/infra/serializers/tracking/signality.py
```diff
@@ -0,0 +1,241 @@
...
+class SignalityInputs(NamedTuple):
+    meta_data: IO[bytes]
+    venue_information: IO[bytes]
+    raw_data_feeds: Iterable[FileLike]
+
+
+class SignalityDeserializer(TrackingDataDeserializer[SignalityInputs]):
+    def __init__(
+        self,
+        limit: Optional[int] = None,
```
probberechts, line 45: It seems `limit` is not implemented.

kloppy/infra/serializers/tracking/signality.py
```diff
@@ -0,0 +1,241 @@
...
+
+        return frame_rate
+
+    @classmethod
+    def __get_periods(cls, raw_data_feeds) -> List[Period]:
+        """gets the Periods contained in the tracking data"""
+        periods = []
+        for period_id, p_tracking in enumerate(raw_data_feeds, 1):
+            first_frame = p_tracking[0]
+            start_time = datetime.utcfromtimestamp(
```
probberechts, line 150: https://blog.ganssle.io/articles/2019/11/utcnow.html

kloppy/infra/serializers/tracking/signality.py
```diff
@@ -0,0 +1,241 @@
...
+            orientation = Orientation.NOT_SET
+
+        metadata = Metadata(
+            teams=teams,
+            periods=periods,
+            frame_rate=frame_rate,
+            orientation=orientation,
+            pitch_dimensions=transformer.get_to_coordinate_system().pitch_dimensions,
+            coordinate_system=transformer.get_to_coordinate_system(),
+            flags=None,
```
probberechts, line 234: The data has "state" and "ball.team". I would assume these define the ball state and ball-owning team?

kloppy/_providers/signality.py
DriesDeprest: This seems to incorrect for all providers, no? Would it be better to just do it the same as for the other providers here and then create a separate PR later on that fixes it for all providers?

PR #358, opened 2024-10-29
[Wyscout v3] Handle unrecognized players gracefully to prevent crashes

DriesDeprest:
In very rare cases Wyscout v3 event data contains an event that is executed by a player that is not listed out in the players of the executing team. The deserializes caused an error in that case. This fix sets the player as None if the player performing the event data is not listed as one of the players of the team performing the event.

### DIFF
--- a/kloppy/infra/serializers/event/wyscout/deserializer_v3.py
+++ b/kloppy/infra/serializers/event/wyscout/deserializer_v3.py
@@ -40,8 +40,8 @@ from kloppy.domain import (
     TakeOnResult,
     Team,
 )
-from kloppy.exceptions import DeserializationError
-from kloppy.utils import performance_logging, DeserializationWarning
+from kloppy.exceptions import DeserializationError, DeserializationWarning
+from kloppy.utils import performance_logging
 
 from ..deserializer import EventDataDeserializer
 from .deserializer_v2 import WyscoutInputs
@@ -786,7 +786,9 @@ class WyscoutDeserializerV3(EventDataDeserializer[WyscoutInputs]):
                 elif player_id not in players[team_id]:
                     player = None
                     warnings.warn(
-                        f"Player {player_id} not listed in the team performing the event.",
+                        f"Event {raw_event['id']} was performed by player {player_id} and team {team_id}, "
+                        f"but the player does not appear to be part of that team's lineup. "
+                        f"Handled by setting the event's player to None.",
                         DeserializationWarning,
                     )
                 else:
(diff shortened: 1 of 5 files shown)


probberechts:
What do you think about raising a `DeserializationWarning` when this happens? It might be helpful to notify the user that there's a potential issue with the data encountered during deserialization.

DriesDeprest:
Does this work as an implementation @probberechts ?

kloppy/tests/prs/pr_358/test_pr_358.py
```diff
@@ -0,0 +1,39 @@
...
+)
+
+from kloppy import wyscout
+
+
+def test_ignore_unknown_player(base_dir):
+    dataset = wyscout.load(
+        event_data=base_dir / "prs" / "pr_358" / "wyscout_events_v3.json",
+        coordinates="wyscout",
+    )
```
probberechts, line 35:
```
with pytest.warns(DeserializationWarning, match="..."):
    ...
```

kloppy/infra/serializers/event/wyscout/deserializer_v3.py
```diff
@@ -780,6 +781,17 @@ def deserialize(self, inputs: WyscoutInputs) -> EventDataset:
...
                     ]
 
+                if player_id == INVALID_PLAYER:
+                    player = None
+                elif player_id not in players[team_id]:
+                    player = None
+                    warnings.warn(
+                        f"Player {player_id} not listed in the team performing the event.",
+                        DeserializationWarning,
+                    )
```
probberechts, line 791: It should be clear which event failed and how it was handled.

```py
warnings.warn(
        f"Event {event_id} was performed by player {player_id} and team {team_id}, "
        f"but the player does not appear to be part of that team's lineup. "
        f"Handled by setting the event's player to None.",
        DeserializationWarning,
    )
```

probberechts (changes requested):

kloppy/tests/prs/pr_358/test_pr_358.py
probberechts: Haha, I didn't expect that this would work 😄 With "..." I actually meant that you should fill this in with the actual warning's text to make sure the correct one is printed.

DriesDeprest: Apologies, I don't have experience with the warnings' module and wasn't paying enough attention 😄 

Should I then just fill in "Event" so it

PR #363, opened 2024-11-12
Add functionality to list StatsBomb open competitions and available matches

UnravelSports:
Hi!

While working on the docs I realized it wasn't very straightforward to see which competitions and matches are available in the StatsBomb open data repository, at least not from within Kloppy.

I have added functionality such that we can do:

```python
from kloppy import statsbomb

statsbomb.list_open_competitions(detailed=True)
statsbomb.list_open_matches(223, 282, detailed=True)
```

A couple things to consider:
- Both these functions return a pandas dataframe, but Kloppy has no explicit dependency on pandas. I have inserted a warning (similar to the one shown when doing `dataset.to_df()` if pandas has not been installed.
- I have added a `detailed` (bool=True) parameter to both functions. Setting it to `False` will provide _all_ available columns, leaving it to the default `True` will give us only the most relevant columns  
- I have moved the underlying functions to `kloppy.infra.serializers.event.statsbomb.helpers`. I'm not sure if this is desirable.
- Because of point 1, we cannot explicitly state the return type of our functions, ie we cannot do the following:
```python
def list_open_matches(
    competition_id: int, season_id: int, detailed: bool = False
) -> pd.DataFrame:
```
Instead we simply do:
```python
def list_open_matches(
    competition_id: int, season_id: int, detailed: bool = False
):
```

### DIFF
--- a/kloppy/infra/serializers/event/statsbomb/helpers.py
+++ b/kloppy/infra/serializers/event/statsbomb/helpers.py
@@ -14,6 +14,18 @@ from kloppy.domain import (
 )
 from kloppy.exceptions import DeserializationError
 
+OPEN_COMPETITIONS_PATH = "https://raw.githubusercontent.com/statsbomb/open-data/master/data/competitions.json"
+OPEN_MATCHES_PATH = "https://raw.githubusercontent.com/statsbomb/open-data/master/data/matches/{competition_id}/{season_id}.json"
+
+import requests as re
+
+
+def get_response(path):
+    response = re.get(path)
+    response.raise_for_status()
+    data = response.json()
+    return data
+
 
 def parse_str_ts(timestamp: str) -> float:
     """Parse a HH:mm:ss string timestamp into number of seconds."""
@@ -141,3 +153,102 @@ def parse_freeze_frame(
         ball_owning_team=event.ball_owning_team,
         other_data={"visible_area": visible_area},
     )
+
+
+def parse_open_matches(competition_id: int, season_id: int, detailed: bool):
+    try:
+        import pandas as pd
+    except ImportError:
+        raise ImportError(
+            "Seems like you don't have pandas installed. Please"
+            " install it using: pip install pandas"
+        )
+    except AttributeError:
+        raise AttributeError(
+            "Seems like you have an older version of pandas installed. Please"
+            " upgrade to at least 1.5 using: pip install pandas>=1.5"
+        )
+
+    path = OPEN_MATCHES_PATH.format(
+        competition_id=competition_id, season_id=season_id
+    )
+    matches = get_response(path)
+    df = pd.json_normalize(matches, sep="_")
+
+    # clean up column names
+    for v in ["competition", "season", "home_team", "away_team"]:
+        df.columns = [x.replace(f"{v}_{v}", v) for x in df.columns]
+
+    if detailed:
+
+        def __flatten_team_managers(df, team_type):
+            # Normalize the team managers, explode to handle list entries within the column, and add prefix for clarity
+            managers_expanded = pd.json_normalize(
+                df[f"{team_type}_managers"].explode(), sep="_"
+            ).add_prefix(f"{team_type}_manager_")
+            return managers_expanded.reset_index(drop=True)
+
+        # Normalize both home and away managers
+        home_managers_expanded = __flatten_team_managers(df, "home_team")
+        away_managers_expanded = __flatten_team_managers(df, "away_team")
+
+        # Concatenate both expanded managers back to the original DataFrame, and drop the original manager columns
+        df = pd.concat(
+            [
+                df.reset_index(drop=True),
+                home_managers_expanded,
+                away_managers_expanded,
+            ],
+            axis=1,
+        )
+        df.drop(
+            columns=["home_team_managers", "away_team_managers"],
+            inplace=True,
+            errors="ignore",
+        )
+        return df
+    else:
+        return df[
+            [
+                "match_id",
+                "match_date",
+                "home_score",
+                "away_score",
+                "home_team_name",
+                "home_team_id",
+                "away_team_name",
+                "away_team_id",
+            ]
+        ]
+
+
+def parse_open_competitions(detailed: bool):
+    try:
+        import pandas as pd
+    except ImportError:
+        raise ImportError(
+            "Seems like you don't have pandas installed. Please"
+            " install it using: pip install pandas"
+        )
+    except AttributeError:
+        raise AttributeError(
+            "Seems like you have an older version of pandas installed. Please"
+            " upgrade to at least 1.5 using: pip install pandas>=1.5"
+        )
+
+    competitions = get_response(OPEN_COMPETITIONS_PATH)
+    df = pd.DataFrame(competitions).sort_values(
+        by="match_available", ascending=False
+    )
+    if detailed:
+        return df
+    else:
+        return df[
+            [
+                "competition_id",
+                "season_id",
+                "country_name",
+                "competition_name",
+                "season_name",
+            ]
+        ]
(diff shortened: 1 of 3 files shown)


UnravelSports:
Just to add, if the basic functionality is approved I'll add some tests

probberechts:
A few remarks:

1. I prefer the initial proposal of adding a `list_open_data` function that immediately retrieves all match ids for all competitions. It can have optional arguments to filter on a competition and / or season. With `list_open_competitions` and `list_open_matches` it's again a two-step operation to retrieve a certain match id.
2. StatsBomb has a great well-maintained Python API. Would it be an option to simply wrap their API? 

```python
def list_open_data(fmt="dataframe"):
    try:
        from statsbombpy import sb
    except ImportError:
        print("Please install the statsbombpy library to use this function.")
        return

    try:
        # Fetch all available competitions as a dictionary
        competitions = sb.competitions(fmt="dict")

        # Initialize an empty list to store DataFrames for matches
        all_matches = []

        # Loop through each competition and fetch matches using competition_id and season_id
        for competition in competitions.values():
            matches = sb.matches(
                competition_id=competition['competition_id'], 
                season_id=competition['season_id'],
                fmt=fmt
            )
            all_matches.append(matches)

        # Combine all matches into a single DataFrame
        if fmt == "dataframe":
            import pandas as pd
            combined_matches = pd.concat(all_matches, ignore_index=True)
            return combined_matches
        elif fmt == "dict":
            return all_matches
        else:
            raise ValueError("Invalid format. Use 'dataframe' or 'dict'.")

    except Exception as e:
        raise RuntimeError(f"An error occurred while fetching data: {e}")

# Example usage:
df = list_open_data(fmt="dataframe")
print(df.head())
```

UnravelSports:
This seems fine to me too! I wasn't sure about relying on the sb package initially, but why not.

UnravelSports:
@probberechts do you have any suggestions for passing the competition / season filters? It might make sense to have this be a dict with comp / season_id value pairs, but it might also make sense to simply be able to provide a competition_id and or a competition / season_id pair?

probberechts:
> do you have any suggestions for passing the competition / season filters?

Maybe simply allow everything? You can define a `StatsBombPartitionIdentifier` object and then try to parse a dict or tuple into this object.

```py
@dataclass
class StatsBombPartitionIdentifier:
    competition_id: Optional[int] = None
    season_id: Optional[int] = None

    @classmethod
    def from_dict(cls, data: Dict):
        """Create a PartitionIdentifier from a dictionary."""
        return cls(
            competition_id=data.get('competition_id'),
            season_id=data.get('season_id')
        )

    @classmethod
    def from_tuple(cls, data: Tuple[Optional[int], Optional[int]]):
        """Create a PartitionIdentifier from a tuple."""
        if len(data) != 2:
            raise ValueError("Tuple must have exactly two elements: (competition_id, season_id)")
        return cls(
            competition_id=data[0],
            season_id=data[1]
        )
 ```
 
Then, the input to `list_open_data` can be a `StatsBombPartitionIdentifier | list[StatsBombPartitionIdentifier] | dict | list[dict] | tuple | list[tuple]`.

Although, I don't think I would use these identifiers a lot. I can see it might be nice to have this option since it will make data retrieval more efficient (you won't have to retrieve the file with matches for each competition-season pair). However, the only ID that I know by hard is the one of the EPL and would probably rather wait an extra second rather than typing another parameter 🙃. So, maybe we don't have to add this option.

>  the first step is always going to be a two step process

Yes, typically it will be. But I prefer to do these two steps directly in Pandas. Below are a few use cases in which I think this functionality would be useful:

:thinking: _I would like to do some analysis with 360 data, which datasets can I use?_

```py
df = list_open_data(fmt="dataframe") 
df["has_360"] = df["match_status_360"] == "available"
print(df.groupby(["competition", "season"])["has_360"].sum().sort_values(ascending=False))  
```

:thinking: _What is the ID of the 2022 World Cup final?_

```py
df = list_open_data(fmt="dataframe") 
print(df[df.match_date == "2022-12-18"]) 
```

:thinking: _Get the metadata of a particular game._

```py
df = list_open_data(fmt="dataframe") 
print(df[df.match_id == 7584]) 
```

UnravelSports:
I've changed it to your suggested implementation. Where we allow a single combination of `competition_id` and `season_id` or we pull everything.

Some notes:
- Have suppressed the `NoAuthWarning` in the `sb.matches()` for-loop because one warning is enough, and we get that warning when we grab the competitions.
- I had to add the `competition_id` and `season_id` to the dataframe implementation, because for some reason statsbomb was not giving that by default.

@koenvo do we need tests for this? And if so, what would they look like?

kloppy/infra/serializers/event/statsbomb/helpers.py
```diff
@@ -14,6 +14,18 @@
 )
 from kloppy.exceptions import DeserializationError
 
+OPEN_COMPETITIONS_PATH = "https://raw.githubusercontent.com/statsbomb/open-data/master/data/competitions.json"
+OPEN_MATCHES_PATH = "https://raw.githubusercontent.com/statsbomb/open-data/master/data/matches/{competition_id}/{season_id}.json"
+
+import requests as re
+
+
+def get_response(path):
```
koenvo, line 23: Please move this to a more generic place like `utils.py`

koenvo:
@probberechts what do you think of putting `ingestify` in between? I might have broken the current statsbomb open data `Source`, but when it's fixed it provides all necessary parts to work with the open data. 

It might need an additional "get or fetch" method to either get it directly from (local) storage, or retrieve it first.

probberechts:
> @probberechts what do you think of putting `ingestify` in between? I might have broken the current statsbomb open data `Source`, but when it's fixed it provides all necessary parts to work with the open data.
> 
> It might need an additional "get or fetch" method to either get it directly from (local) storage, or retrieve it first.

I think adding ingestify might be overkill for our needs. We only need to download ~10 small JSON files, maybe once in a while. The main reason I proposed using the statsbombpy package is that it already parses everything neatly into dataframes. I am not very familiar with ingestify, but I guess it does not do that part.

UnravelSports:
I also don't know what ingestify does, but does indeed feel like overkill. The statsbombpy package works fine, plus it allows users to directly grab non open data too if they have access

koenvo:
Ok. Let’s go for this solution. 

I do think it would be good to introduce some models for competition, seasons and match etc to prevent having StatsBomb specific structures in through the code.

UnravelSports:
How/where do you see this?

probberechts:
> I do think it would be good to introduce some models for competition, seasons and match etc to prevent having StatsBomb specific structures in through the code.

Yes, this is a good point! I guess what you mean is that we should add something like a `DatasetCollection` entity to kloppy?

```py
@dataclass
class Game:
    game_id: int
    home_team: Team
    away_team: Team
    date: datetime

@dataclass
class Season:
    season_year: str  # e.g., "2023-2024"
    season_id: int
    games: List[Game] = field(default_factory=list)

@dataclass
class Competition:
    name: str  # e.g., "Premier League"
    competition_id: int
    seasons: List[Season] = field(default_factory=list)

@dataclass
class DatasetCollection:
    """Represents a dataset containing games from multiple competitions."""
    competitions: List[Competition] = field(default_factory=list)
        
   def to_df():
       """Retrieve all games from all competitions and seasons as a dataframe."""
       raise NotImplementedException()     
 ```

UnravelSports:
@probberechts I created this to be able to more easily write the docs (aka, how to load stats bomb open data). However, now I'm not quite sure about the status of this PR. Do you want to have `DatasetCollection` in place before merging? Or could we potentially do that at a later stage?

UnravelSports:
@koenvo any updates on this? I was checking the docs and noticed this is still open.

PR #373, opened 2024-12-14
Fix common bug in parsing of UTC datetimes; remove dateutil

probberechts:
This PR addresses a common bug related to parsing UTC datetimes, fixes the timezone of Opta F24 data and removes the `dateutil` dependency.

The following conventions are used:
- All datetimes are timezone-aware
- If the local timezone is known, the local timezone is used
- If the local timezone is not known, UTC is used

## `astimezone` -> `replace`
Often, datetimes are specified as a string and use the UTC timezone. In kloppy, we want to parse these a timezone-aware `datetime`. Then, then string was usually parsed and converted to a timezone-aware datetime as follows:

```py
>>> parse("2024-12-14 12:00").astimezone(timezone.utc)
datetime.datetime(2024, 12, 14, 11, 0, tzinfo=datetime.timezone.utc)  # on a system with local timezone UTC+1
```

The correct implementation is:

```py
>>> parse("2024-12-14 12:00").replace(tzinfo=timezone.utc)
datetime.datetime(2024, 12, 14, 12, 0, tzinfo=datetime.timezone.utc)
```

The former _changes_ the timezone, while the latter _sets_ the timezone (i.e., makes the datetime timezone-aware).

## Opta F24 timezone

According to the documentation, timestamps in the F24 files use the timezone "Europe/London".

## Remove `dateutil`

We know the exact format of the date string in advance. Hence, we don't need `dateutil`'s `parser` module.

### DIFF
--- a/kloppy/infra/serializers/event/datafactory/deserializer.py
+++ b/kloppy/infra/serializers/event/datafactory/deserializer.py
@@ -1,9 +1,8 @@
 import json
 import logging
-from datetime import timedelta, datetime, timezone
-from dateutil.parser import parse, _parser
 from dataclasses import replace
-from typing import Dict, List, Tuple, Union, IO, NamedTuple
+from datetime import datetime, timedelta, timezone
+from typing import IO, Dict, List, NamedTuple, Tuple, Union
 
 from kloppy.domain import (
     AttackingDirection,
@@ -40,7 +39,6 @@ from kloppy.domain import (
 from kloppy.exceptions import DeserializationError
 from kloppy.infra.serializers.event.deserializer import EventDataDeserializer
 from kloppy.utils import Readable, performance_logging
-
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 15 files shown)


probberechts:
Is there a one-minute difference between UK/London and UTC? Or did I make a mistake?

```py
import datetime
from pytz import timezone

naive_datetime = datetime.datetime.now()
local_datetime = naive_datetime.replace(tzinfo=timezone('Europe/London'))
utc_datetime = local_datetime.astimezone(timezone('UTC'))
print("Naive datetime:", naive_datetime)
print("Local datetime:", local_datetime)
print("UTC datetime:  ", utc_datetime)
```

This is the output:

```
Naive datetime: 2024-12-14 21:53:04.383850
Local datetime: 2024-12-14 21:53:04.383850-00:01
UTC datetime:   2024-12-14 21:54:04.383850+00:00
```

PR #374, opened 2024-12-17
[General] Fixed Undersirable "None" value for positions.last() 

UnravelSports:
Hi,

I noticed that with the new `TimeContainer` functionality, when we call `player.positions.last()` we would get a `None` value as the last position for a lot of players. I'm assuming this is because of substitutions? 

I've resolved the issue by now simply returning the last not None value, such that at least for every player that was on the pitch during the game the `last()` call gives a valid position.

Additionally, since I was getting a deprecation warning I fixed some very small syntax in the f24 xml parsing. This is not part of the same problem though.

### DIFF
--- a/kloppy/domain/models/time.py
+++ b/kloppy/domain/models/time.py
@@ -241,13 +241,20 @@ class TimeContainer(Generic[T]):
         return ranges_
 
     def last(self, include_time: bool = False, default=SENTINEL):
+        def __get_last_non_none_key(items):
+            for key, value in reversed(items.items()):
+                if value is not None:
+                    return key
+            return None
+
         if not len(self.items):
             if default == SENTINEL:
                 raise KeyError
             else:
                 return default
 
-        time = self.items.keys()[-1]
+        time = __get_last_non_none_key(self.items)
+
         if include_time:
             return time, self.items[time]
         else:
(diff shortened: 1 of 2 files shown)


koenvo:
The substitutions are an issue indeed. I would like to have a way to get the last value (including `None` in case the player is substituted). Maybe add a flag/argument to enable the new behaviour?

UnravelSports:
It's been a while, you mean add an argument to the `.last()` call? Something like `include_none`?

UnravelSports:
@koenvo should be fixed now

kloppy/domain/models/time.py
```diff
@@ -239,14 +239,29 @@ def ranges(self) -> List[Tuple[Time, Time, T]]:
...
             if default == SENTINEL:
                 raise KeyError
             else:
                 return default
 
-        time = self.items.keys()[-1]
+        if include_none:
+            time = self.items.keys()[-1]
+        else:
+            time = __get_last_non_none_key(self.items)
```
koenvo, line 263: This will break when there a only `None` values in `self.items`, correct? What should happen when `time` get a value of `None`? Maybe similar to https://github.com/PySport/kloppy/pull/374/files#diff-634d7febc3eb18e4cadff92f381df2b9125bec36feef3cea5e08762b4fab9d3cR255

UnravelSports: I have to admit I don't quite follow what you're saying. 

If self.items.keys()[-1] returns `None` we have a different problem right? Because that means some how, somewhere a position change at time None was introduced. Is that possible right now?

Anyway, if we want to deal with that than what would be the expected return value of `last()` if `time = None`?

Would this make sense?

```python
if include_time and time is not None:
            return time, self.items[time]
        elif not include_time:
            return self.items[time]
        else:
            return None
```

PR #377, opened 2024-12-18
Refactor tracking data model

probberechts:
This pull request proposes a new domain model for tracking data and updates various serializers to align with it.

The idea is that a tracking dataset is a collection of object detections. There are two ways to organize this collection:

1. As a sequence of frames where each frame contains the coordinates of all object that were detected at a point in time
2. As a set of trajectories where each trajectory contains the coordinates of a single object over a period of time.

uml_tracking

The deserializers still create the frame-based representation but you can now also compute the trajectory-based representation for any trackable object.

## Why this is better
I think this has a few advantages:
1. Which representation is most convenient depends on the use case. A frame-based representation allows to analyze player interactions, while a trajectory-based representation allows to analyze a player's kinematics. Now, we allow both representations.
2. Previously we had different attributes and logic for handling ball coordinates and player coordinates. Now both the ball and the ball and players are handled as "trackable objects". It also becomes easier to implement support for tracking other objects (e.g., the referee).
3. I need this for #298 

## Breaking changes
- the `PlayerData` entity was replaced by a `Detection` entity

## Still thinking about

- How can we best support joint tracking data? A subclass of `Detection`, a `Detection.joint_coordinates` attribute, ...?
- ~Maybe `Detection` is too ambiguous for an entity name?~ *Edit*: Now using `TrackedObjectState`


UnravelSports:
@probberechts when considering joint tracking I would propose not adding it directly as `Detection.joint_coordinates` but perhaps as `Detection.joints`. Where the `joints` attribute can hold not only the coordinates directly, but perhaps a hierarchical description of the joint connections (perhaps as a graph) and potentially derived values like joint angles, or even inverse kinematics (although this is a bit far fetched). 

Additionally, we should probably consider that joint data does not always come in the form of world coordinates, but sometimes it is provided as "angles" (e.g. "head_angle", "hip_angle" etc.) without the inclusion of coordinates.

mifuha:
I came across a `ball_speed` preservation issue during a local investigation of tracking transformations on `master` at commit `51dbd38c4fb48c0815119e23ff9a3a68ea06be52`.

Using the existing Sportec fixtures with `coordinates="sportec"` and `only_alive=False`:

```python
frame = dataset.get_record_by_id(10001)
assert frame.ball_speed == 65.59

transformed = dataset.transform(to_orientation="AWAY_HOME")
assert transformed.get_record_by_id(10001).ball_speed == 65.59
# Fails on the tested commit: actual value is None.
```

The coordinates flip correctly, but `ball_speed` is lost when the frame is rebuilt. Adding `ball_speed=frame.ball_speed` in `__flip_frame` fixes the regression.

The same focused tests go from 2 failed / 2 passed to 4 passed; 78 nearby existing tests also pass with the one-line change. I can share the complete regression test.

I noticed this PR’s new tracking model appears to preserve speed as part of the tracked-object state, so the refactor may already address this. Would a small standalone fix and regression test for the current model be useful, or would you prefer to include the regression here?

PR #386, opened 2024-12-22
Freeze frame position loading

AndrewRook:
Hi there! 

I was trying to use freeze frame data from statsbomb and I think I encountered a couple of small bugs in how player information was connected to the player tracking data. First, there seems to be an issue when getting player positions if a player hasn't been subbed in yet, which I resolved by catching the error and just continuing. Then there's a timing issue when deserializing the data, where position information doesn't seem to be added until the `EventDataset` is created. I resolved that issue by postponing setting the freeze frame data until after the `EventDataset` is initialized.

I wasn't sure exactly how the unit tests are laid out, so I added a test for each apparent bug in the `test_statsbomb` test module, let me know if there's a better place for them. Also, I have no idea if this is the right way to solve these issues (or whether it's about to be obviated by the changes in #377), so please feel free to let me know if there's another way to do it or just close the PR; I just figured I should share my workarounds for the problems I had in case they're more broadly helpful.

### DIFF
--- a/kloppy/domain/models/common.py
+++ b/kloppy/domain/models/common.py
@@ -264,7 +264,10 @@ class Team:
     def get_player_by_position(self, position: PositionType, time: Time):
         for player in self.players:
             if player.positions.items:
-                player_position = player.positions.value_at(time)
+                try:
+                    player_position = player.positions.value_at(time)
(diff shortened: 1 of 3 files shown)


probberechts:
Thanks!

Issue #388, opened 2024-12-25
By using add_state, the positions and formations information gets incorrectly "double" updated


DriesDeprest:
**Description**
The `add_state()` method ends by replacing the records of a dataset with records that have the relevant state(s), as a result of `replace(dataset, records=events)`. This `replace` method of the dataclasses library, will "recreate" the `Dataset` and thus re-trigger the `__post_init` logic, including `_init_player_positions()` and `_update_formations_and_positions()`. As a result, positions and formations might be incorrect.

**How to reproduce**
When e.g. running the `test_score_state_builder()` in `test_state_builder.py`. The positions of Philippe Coutinho (seen in `dataset.metadata.teams[0].players[1].positions.items`) will be different before and after the `add_state()` logic is applied.

**Solution**
Modify the Dataset class to include a flag that determines whether the additional initialization logic runs. For example:

```python
@dataclass
class Dataset(ABC, Generic[T]):
    """
    Dataset

    Attributes:
        records:
        metadata: Metadata for this Dataset
    """

    Column = NewType("Column", Union[str, Callable[[T], Any]])

    records: List[T]
    metadata: Metadata
    skip_init: bool = False  # New flag to control initialization logic

    def __iter__(self):
        return iter(self.records)

    def __getitem__(self, item):
        return self.records[item]

    def __len__(self):
        return len(self.records)

    def __post_init__(self):
        if self.skip_init:
            return  # Skip further initialization if skip_init is True

        for i, record in enumerate(self.records):
            record.set_refs(
                dataset=self,
                prev=self.records[i - 1] if i > 0 else None,
                next_=self.records[i + 1] if i + 1 < len(self.records) else None,
            )

        self._init_player_positions()
        self._update_formations_and_positions()

```

and pass the `skip_init` to the `add_state()` function

```python
return replace(dataset, records=events, skip_init=True)
```

probberechts:
This is a great catch! However, I’m not entirely convinced by the proposed solution. It adds complexity to the initialization logic without fully addressing the core issue.

Consider this scenario: you identify an error in the timing of a substitution and fix it by updating the dataset. In such a case, you’d need to set `skip_init=False`, which would lead to incorrect positions if my understanding of the issue is correct. 

Instead, I’d recommend resetting the positions before `_init_player_positions` runs. 

Do you also happen to know why this is currently failing? While I can see how altering event timings might cause issues, I haven’t yet identified the exact source of the problem.

DriesDeprest:
[...]

PR #389, opened 2024-12-26
Reset positions and formations during `EventDataset` creation

DriesDeprest:
Fix for issue https://github.com/PySport/kloppy/issues/388.

### DIFF
--- a/kloppy/domain/models/common.py
+++ b/kloppy/domain/models/common.py
@@ -220,9 +220,6 @@ class Player:
     def set_position(self, time: Time, position: Optional[PositionType]):
         self.positions.set(time, position)
 
-    def reset_positions(self):
-        self.positions.__init__()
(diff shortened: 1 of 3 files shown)


kloppy/domain/models/common.py
```diff
@@ -220,6 +220,9 @@ def __eq__(self, other):
     def set_position(self, time: Time, position: Optional[PositionType]):
         self.positions.set(time, position)
 
+    def reset_positions(self):
+        self.positions.__init__()
```
probberechts, line 224: I don't think directly calling `__init__()` is a good idea. I was confused by it.

Can you add a `reset` method to the `Timecontainer` class instead?

```py
    def reset(self):
        """Clears all items from the container."""
        self.items.clear()
  ```

PR #390, opened 2024-12-26
[Wyscout v3] Deduce kick-off qualifier

### LINKED ISSUE #360: [Wyscout v3] Deduce kick off set piece type for events
Wyscout does not label kick-off events with a "kick-off" tag. 

We could deduce what events are kick-offs by assuming:
1. The first pass event of any period is a kick-off
2. The first pass event after a goal is a kick-off

Thoughts?

rmotmans:
For Wyscout we check for pass events if it is the first event in a period or if it happens after a goal. In these cases the pass event is also the kick-off.

Closes #360

### DIFF
--- a/kloppy/infra/serializers/event/wyscout/deserializer_v3.py
+++ b/kloppy/infra/serializers/event/wyscout/deserializer_v3.py
@@ -772,6 +772,7 @@ class WyscoutDeserializerV3(EventDataDeserializer[WyscoutInputs]):
 
             events = []
 
+            next_pass_is_kickoff = False
             for idx, raw_event in enumerate(raw_events["events"]):
                 next_event = None
                 next_period_id = None
@@ -780,6 +781,14 @@ class WyscoutDeserializerV3(EventDataDeserializer[WyscoutInputs]):
                     next_period_id = _parse_period_id(
                         next_event["matchPeriod"]
                     )
+
+                if (
+                    idx == 0
+                    or raw_event["matchPeriod"]
+                    != raw_events["events"][idx - 1]["matchPeriod"]
+                    or "conceded_goal" in raw_event["type"]["secondary"]
+                ):
+                    next_pass_is_kickoff = True
 
                 team_id = str(raw_event["team"]["id"])
                 team = teams[team_id]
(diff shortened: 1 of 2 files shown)


probberechts:
Thanks for this contribution!

I think it is a bit naive to assume that the kick off will always be the first event after a goal. For example, it can happen that a player gets a card after the goal for taking off their shirt during the celebration. 

I would probably implement this with a flag `next_pass_is_kickoff` and set it to `True` after a goal. What do you think?

rmotmans:
That was actually my first attempt but I saw some edge cases where a goal would be disallowed. And i didn't see a clear tag for a disallowed goal to handle this properly. 
In one example I saw as consecutive events Goal -> whistle -> free kick -> pass, and then that pass would be considered a kick-off which would be wrong. 
I'll look a bit deeper tomorrow how to handle both cases simultaneously.

DriesDeprest:
@rmotmans can you clarify where you found an example of a disallowed goal in the raw Wyscout V3 event data?

Based on my experience and what I read in the [data glossary](https://dataglossary.wyscout.com/goal/) I would think they would not add disallowed goals to the raw Wyscout V3 event data. They don't really mention it explicitly, so I could be wrong on this.

If disallowed goals do appear in the Wyscout V3 event data as a shot event with a secondary event type goal, then we would have a bigger problem as these would currently be incorrectly recognized as allowed goals 😅 

For the sake of recognizing kick off events, I think the best option would be to work with a flag but set it based on a [Goal conceded](https://dataglossary.wyscout.com/goal_conceded/) event, since I'm pretty sure they will not add this to the raw event data for disallowed goals and this event will follow both after a goal and an [own goal](https://dataglossary.wyscout.com/own_goal/), which are two different events for Wyscout. I see that you are doing this for V3 already, but not for V2. 

I also noticed that we do not handle own goals in our Wyscout V3 serializer properly, so I'll create a PR for recognizing this properly.

DriesDeprest:
FYI; here is the [PR](https://github.com/PySport/kloppy/pull/393) for the recognizing own goal events.

DriesDeprest:
I checked with @rmotmans, and it seems that disallowed goals are indeed added to the event stream for Wyscout V2. I've added a dedicated issue for this: https://github.com/PySport/kloppy/issues/394.

Robbe and I suggest that we limit the scope of this PR to deducing kick-off events for Wyscout V3 using the conceded goal event and a `next_pass_is_kickoff` flag. 

Once https://github.com/PySport/kloppy/issues/394 is fixed, someone could then create a PR for deducing kick-off events for Wyscout V2, if desired.

Wdyt @probberechts ?

rmotmans:
Implemented only for Wyscout V3 like discussed with @DriesDeprest

probberechts:
Ok! Could you also add a test?

Issue #249, opened 2023-12-04
[Stats Perform] Deduce receiving player information for Pass event 


DriesDeprest:
Unlike other data providers, Stats Perform does not give information about the pass receiver in the raw data of their pass events.

We could, however, deduce the receiving player ID from the next event. If we consider the next event after a successful pass and use the ID of the player executing that event, if the next event is a ball owning event.

@koenvo do you think this is a good idea?

probberechts:
Yes, this is a good idea. If you can *reliably infer* who is the intendend recipient of a pass, you can impute the `receiver_player`. It probably requires a few more checks though (e.g., filter out deflected passes, passes interrupted by a foul, etc.).

Issue #265, opened 2023-12-26
receiver_coordinates naming


koenvo:
The `receiver_coordinates` property of the `PassEvent` has a naming issue. Only in case of a sucessful received pass the naming makes sense. A more generic name could be `end_coordinates`.

PR #392, opened 2024-12-27
[Stats Perform] Identifying the pass receiver on event deserialization

douglaspampu:
Change related to the issue: https://github.com/PySport/kloppy/issues/249

Setting the receiver_player and the receive_timestamp if the pass result is complete and the on the next event is in BALL_OWNING_EVENTS and the team with the ball possession is the same.

This way I am filtering intercepted passes that may lead to a corner, throw in or a fault. Events that would not change the team with ball possession, but the event player would not be the pass receiver. On this case the pass receiver is the player that after the pass event, actually does something with the ball.


probberechts:
At first sight, this looks good to me. Thanks! Would you also be able to add a few test cases?

douglaspampu:
@probberechts I added a test for the ball receiver checking the base cases, the pass is complete and the team kept the ball, the pass is complete but there was duel and the ball possession changed after the pass.

For the test I followed the approach of the incomplete pass test, but it is depending on the to_pandas function to run, and I see this function will be deprecated. I also noted that the TestStatsPerformEvent tests are not running (this would be the best place for these tests). I appreciate any inputs on the these tests placement, if its ok like this or if I should look closely on the TestStatsPerformEvent not running

Thanks

DriesDeprest:
Hi Douglas,

I think the test for checking the pass receiver for the Stats Perform deserializer should indeed be part of the `TestStatsPerformEvent` class in the `test_statsperform.py` file. It doesn't really make sense to add it to specific test cases in the `test_helpers.py` whose main goal is to check how Event- or TrackingDataset's are converted to dataframes. 

I would probably break refactor the class to have multiple test functions to make it more readable and maintainable:
```python
class TestStatsPerformEvent:
    """Tests related to deserializing the MA3 event data feed.

    See Also:
        kloppy.tests.test_opta.TestOptaEvent
    """

    def test_all_events(self, event_dataset: EventDataset):
        assert len(event_dataset.records) == 1643

    def test_substitutions(self, event_dataset: EventDataset):
        substitution_events = event_dataset.find_all("substitution")
        assert len(substitution_events) == 9

        m_wintzheimer = event_dataset.metadata.teams[0].get_player_by_id(
            "aksjicf4keobpav3tuujngell"
        )
        b_jatta = event_dataset.metadata.teams[0].get_player_by_id(
            "3mp7p8tytgkbwi8itxl5mfkrt"
        )

        first_sub = substitution_events[0]

        assert first_sub.time == Time(
            period=event_dataset.metadata.periods[1],
            timestamp=timedelta(seconds=946, microseconds=475000),
        )
        assert first_sub.player == m_wintzheimer
        assert first_sub.replacement_player == b_jatta
        
    def test_passes(self, event_dataset: EventDataset):
        pass_events = event_dataset.find_all("pass")
        assert len(substitution_events) == X.   # replace X with the actual amount of passes in the dataset
        
        # here comes your test for checking the pass receiver of a given pass
    

```
and the assertion statement for the coordinate system I would move to the `TestStatsPerformMetadata` class.
```python
class TestStatsPerformMetadata:
    """Tests related to deserializing the MA1 meta data feed."""

    def test_provider(self, tracking_dataset: TrackingDataset):
        assert tracking_dataset.metadata.provider == Provider.STATSPERFORM
        
    def test_orientation(self, tracking_dataset: TrackingDataset):
            assert event_dataset.metadata.coordinate_system == OptaCoordinateSystem(
                  # StatsPerform does not provide pitch dimensions
                  pitch_length=None,
                  pitch_width=None,
              )
```

Another option would be to add it to append a test function to the `TestOptaPassEvent` class in the `test_opta.py` file, which checks the pass receiver for a given pass. 

What do you mean when you say that the TestStatsPerformEvent tests are not running? Do you mean that the Github Action Workflow is failing and not running the tests? If so, this is a known issue which is the result of an issue with one of the dependencies of the docs. See: https://github.com/PySport/kloppy/pull/356#issuecomment-2566466256 for reference. I would suggest to just run the tests locally for now to check your PR code changes 


**For your context**
The StatsPerform provider is a bit of an edge case since they provide their data in the different file types (.xml / .json) using different "systems" (Opta / Stats Perform).  Opta is their legacy system, while their Stats Perform is the more recent counterpart. Since the data is in essence the same regardless of the file type or system used, we first transform the raw events to a list of `OptaEvent` and then serialize it with the `StatsPerformDeserializer`. At some point I would refactor the `test_statsperform.py` and `test_opta.py` into a single file which would test the following:
- correct serialization of raw Opta F7 & 24 xml and StatsPerform MA1 & MA3 xml or json into a stream of OptaEvents + meta data
- correct conversion of a stream of OptaEvents + meta data into an EventDataset
- correct serialization of Stats Perform tracking data
This way, we don't have separate / duplicate tests on the resulting EventDataset for different input "systems". However, this rework is clearly out of scope for this issue/PR and should thus be handled in a separate and dedicated issue/PR imo.

douglaspampu:
@DriesDeprest Thanks for the input. I was having an local issue with my unit tests and not all of them were running, It is solved now.

I made the change you suggested, at least for me, it wasnt out of the scope for this task since it is related to the unit tests for the pass receiver logic

probberechts:
Thanks @douglaspampu. I've refactored the tests and I found one case for which it does not work yet: a successfull pass is sometimes challenged by a defender, in which case the next event is not by a player from the same team. Below is an example: 
```py
challenged_pass = event_dataset.get_event_by_id("2328600271")
assert challenged_pass.receiver_player.player_id == "ci4pwzieoc94uj3i1371bsatx"
```
Do you think it would be possible to support this?

douglaspampu:
Hi @probberechts it is possible. I assumed this case it would not be clear if the pass was aiming the player that received the ball, so I let this case out and focused just on the cases a direct pass found its target.

But I'll add support for this one and the test case

Thanks

douglaspampu:
@probberechts I updated the commit with some changes.

The deflected pass scenario you implemented was the expected scenario for the interrupted pass:
- There is a pass;
- The pass is blocked by a ball touch event with a outcome as 1
- The next event the ball is back in possession of the original team.

But, in this case, the ball turned out the same player that originated the play, which would be weird to be considered a pass and even more if we set this player to be the pass receiver. But not all wrong if there was a pass and he ended up with the ball.

On the challenge pass scenario you proposed, the block event outcome is 0 this made me wonder if the deflect event outcome should be checked at this point. But if we ignore the outcome at all, we could end up with scenarios where the block is successful but the next event the attacking team makes pressure, recover the ball and this counts as a pass.

In my opinion these corner cases would not exactly qualify as a pass. On the current state of the change we are covered by a simple pass deflection and it may be risky adding more scenarios.

douglaspampu:
@probberechts So, to be sure, if a pass is deflected by a ball touch, we would mark it as a complete pass and set the receiver_coordinates as the original implementation did. but the pass receiver would be set up only by a challenge event?
This means, when testing, even if the pass result has a complete status it may not have a pass receiver information if the pass is blocked by a ball touch event.

I added a new check for challenge passes and then the test case you suggested works as expected.

probberechts:
> So, to be sure, if a pass is deflected by a ball touch, we would mark it as a complete pass and set the receiver_coordinates as the original implementation did. but the pass receiver would be set up only by a challenge event?
This means, when testing, even if the pass result has a complete status it may not have a pass receiver information if the pass is blocked by a ball touch event.

Exactly. We've defined these attributes as follows:

- `receiver_coordinates`: The coordinates where the pass was received. (note: #265)
- `receiver_player`: The intended receiver of the pass.

Hence, the `receiver_player` is not necessarily the player that eventually received the pass.

douglaspampu:
So now the PR is meeting these requirements

Thanks for the clarifications @probberechts

kloppy/infra/serializers/event/statsperform/deserializer.py
```diff
@@ -245,39 +247,72 @@
...
+        ):
+            ball_receipt_event = next_next_event
+    # Assign the receiver_player and receive_timestamp
+    # if a ball_receipt_event was found
+    if (
+        ball_receipt_event is not None
+        and ball_receipt_event.player_id is not None
+    ):
+        receiver_player = team.get_player_by_id(ball_receipt_event.player_id)
+        receive_timestamp = ball_receipt_event.timestamp
```
lodevt, line 290: Hi, I noticed a small mistake in this line. The receive timestamp should be a timedelta, relative to the start of the period. In this implementation, the receive_timestamp it not a timedelta but the actual date and exact time when the event took place. 

To fix this, the current period could be passed as an extra argument to the _parse_pass function and the receive_timestamp would become 'ball_receipt_event.timestamp - period.start_timestamp'.

Issue #379, opened 2024-12-19
Deduce synthetic carry events for event providers which don't annotate them


DriesDeprest:
Some event providers, e.g. Stats Perform, do not annotate carry events. However, incorporating carry events can be valuable for downstream analysis.

Would it make sense to build some method for an EventDataset that lacks carry events, which contains logic that analyzes the stream of events and insert carry events at the appropriate points within the dataset?

probberechts:
Yes, this could be useful, but based on my experience with SPADL, it is not possible to achieve this with 100% reliability. This could be a helper method and/or option during the deserialization, but I would not do it by default during the deserialization.

PR #396, opened 2025-01-07
Deduce synthetic carry events for event providers which don't annotate them 

lodevt:
Addressing Issue [#379](https://github.com/PySport/kloppy/issues/379)

I tried to follow a similar approach as in socceraction to add carry events for providers which don't annotate them: (https://github.com/ML-KULeuven/socceraction/blob/af297a490fc521d27622b947e93fa15c5e0092da/socceraction/spadl/base.py#L38).

The implementation and tests are not complete yet. I think that a great way to test this is to remove the carry events from a StatsBomb dataset and then compare the generated carries with the carries from the original dataset. It is impossible to have an accuracy of 100% but still the output should have high similarities.

I am happy to hear any feedback while further implementing this.

### DIFF
--- a/kloppy/infra/serializers/tracking/secondspectrum.py
+++ b/kloppy/infra/serializers/tracking/secondspectrum.py
@@ -155,8 +155,8 @@ class SecondSpectrumDeserializer(
                     first_byte + inputs.meta_data.read()
                 ).match
                 frame_rate = int(match.attrib["iFrameRateFps"])
-                pitch_size_height = float(match.attrib["fPitchYSizeMeters"])
-                pitch_size_width = float(match.attrib["fPitchXSizeMeters"])
+                pitch_size_height = float(match.attrib["fPitchXSizeMeters"])
+                pitch_size_width = float(match.attrib["fPitchYSizeMeters"])
 
                 periods = []
                 for period in match.iterchildren(tag="period"):
@@ -243,7 +243,7 @@ class SecondSpectrumDeserializer(
         # Handles the tracking frame data
         with performance_logging("Loading data", logger=logger):
             transformer = self.get_transformer(
-                pitch_length=pitch_size_width, pitch_width=pitch_size_height
+                pitch_length=pitch_size_height, pitch_width=pitch_size_width
             )
 
             def _iter():


kloppy/domain/models/event.py
```diff
@@ -1186,6 +1186,17 @@ def aggregate(self, type_: str, **aggregator_kwargs) -> List[Any]:
 
         return aggregator.aggregate(self)
 
+    def add_deduced_event(self, event_type_: EventType):
+        if event_type_ == EventType.CARRY:
+            from kloppy.domain.services.event_deducers.carry import (
+                CarryDeducer,
+            )
+
+            deducer = CarryDeducer()
```
probberechts, line 1195: Maybe we can allow passing `**kwargs` to the method such that the `CarryDeducer` can be configured.

kloppy/domain/services/event_deducers/carry.py
```diff
@@ -0,0 +1,142 @@
...
+from kloppy.domain.services.event_deducers.event_deducer import (
+    EventDatasetDeduducer,
+)
+
+
+class CarryDeducer(EventDatasetDeduducer):
+    min_carry_length_meters = 3
+    max_carry_length_meters = 60
+    max_carry_duration = timedelta(seconds=10)
+    event_factory = EventFactory()
```
probberechts, line 23: Use the event factory from the `EventDataDeserializer`.

kloppy/domain/services/event_deducers/carry.py
```diff
@@ -0,0 +1,142 @@
...
+
+    def deduce(self, dataset: EventDataset):
```
probberechts, line 25: Again, I'm not a big fan of the term "deduce".

kloppy/domain/services/event_deducers/carry.py
```diff
@@ -0,0 +1,142 @@
...
+                else:
+                    generic_next_event = False
+                if not event.team.team_id == next_event.team.team_id:
+                    continue
+
+                if next_event.event_type not in valid_event_types:
+                    continue
+                # not headed shot
+                if (
+                    (hasattr(next_event, "body_part"))
```
probberechts, line 65: An `Event` does never have a "body_part" attribute. The body part is added as a qualifier.

kloppy/domain/services/event_deducers/carry.py
```diff
@@ -0,0 +1,142 @@
...
+                new_carry = self.event_factory.build_carry(
+                    **carry_event_args, **generic_event_args
+                )
+                new_carries.append(new_carry)
+
+        for new_carry in new_carries:
+            pos = bisect.bisect_left(
+                [e.time for e in dataset.events], new_carry.time
+            )
+            dataset.records.insert(pos, new_carry)
```
probberechts, line 142: Instead of inserting the generated carries based on time, can't we insert them based on `idx` in the previous loop?

kloppy/domain/services/event_deducers/carry.py
```diff
@@ -0,0 +1,142 @@
...
+)
+from kloppy.domain.services.event_deducers.event_deducer import (
+    EventDatasetDeduducer,
+)
+
+
+class CarryDeducer(EventDatasetDeduducer):
+    min_carry_length_meters = 3
+    max_carry_length_meters = 60
+    max_carry_duration = timedelta(seconds=10)
```
probberechts, line 22: Inlcuding `carry` in these variables is redundant.

kloppy/domain/services/event_deducers/carry.py
```diff
@@ -0,0 +1,142 @@
...
+                        seconds=0.1
+                    )
+                else:
+                    last_timestamp = (
+                        event.timestamp
+                        + (next_event.timestamp - event.timestamp) / 10
+                    )
+
+                generic_event_args = {
+                    "event_id": f"{str(uuid.uuid4())}",
```
probberechts, line 117: We've adopted the convention of adding the prefix "synthetic-" for generated events.

kloppy/domain/services/event_deducers/carry.py
```diff
@@ -0,0 +1,142 @@
...
+                    "coordinates": last_coord,
+                    "team": next_event.team,
+                    "player": next_event.player,
+                    "ball_owning_team": next_event.ball_owning_team,
+                    "ball_state": event.ball_state,
+                    "period": next_event.period,
+                    "timestamp": last_timestamp,
+                    "raw_event": {},
```
probberechts, line 125:
```suggestion
                    "raw_event": None,
```

kloppy/domain/services/event_deducers/carry.py
lodevt: The event factory from the 'EventDataDeserializer' is not directly available in the 'EventDataset'. How would you handle this? Would you pass the event factory when initialising  the 'EventDataset' or would you do this in another way?

probberechts: Is it an option to add a `event_factory` parameter to the `SyntheticEventGenerator`, similar to how it is done for the `EventDataDeserializer`?

lodevt: Yes, I have now implemented it this way.

kloppy/domain/models/event.py
lodevt: Which configuration arguments other than the event_factory would be useful to pass here?

probberechts: I guess all configuration attributes of the `CarryDeducer` like `min_carry_length_meters`, `max_carry_length_meters`, etc.

lodevt: The `SyntheticCarryGenerator` can now be initialised with these attributes.

lodevt:
What do you think would be the best convention to follow to generate the event_ids of the synthetic carries?

The StatsBomb deserialiser follows the convention '<event_type>-<previous_event_id>' (see [duels](https://github.com/PySport/kloppy/blob/e507eb5ebefa8cd2c0471de15b5a3c7407650745/kloppy/infra/serializers/event/statsbomb/specification.py#L338) and [ball out](https://github.com/PySport/kloppy/blob/e507eb5ebefa8cd2c0471de15b5a3c7407650745/kloppy/infra/serializers/event/statsbomb/specification.py#L357C19-L357C22)) and the Wyscout V3 deserializer similarly uses 'synthetic-<team_id>-<previous_event_id>' for synthetic formation change events (see [formation change](https://github.com/PySport/kloppy/blob/e507eb5ebefa8cd2c0471de15b5a3c7407650745/kloppy/infra/serializers/event/wyscout/deserializer_v3.py#L1003)).

Based on this, I suggest using 'carry-<previous_event_id>' as the event_id for the synthetic carries.

Issue #253, opened 2023-12-07
Create OWN_TEAM Orientation


DriesDeprest:
I was thinking about adding support for an `OWN_TEAM` `Orientation` which in the case for an `EventDataset` would behave the same as the `ACTION_EXECUTING_TEAM` `Orientation`, but in the case of a `TrackingDataset` it would orient the coordinates of each player with respect to its own side. Implying that e.g. both goalkeepers would have a starting position of +- (0, 50). 

What do you think @koenvo @JanVanHaaren ?

koenvo:
Sounds interesting. I'm planning to merge https://github.com/PySport/kloppy/pull/190 first.

Curious what @probberechts thinks about your suggestions.

probberechts:
That sounds like a valuable feature! However, I would categorize it more as a symmetry transformation rather than an orientation change. In my conceptualization, orientation transforms merely switch perspectives without altering the underlying data, similar to changing the camera angle. On the other hand, symmetry transforms modify the relationships between data points. 

So maybe we should create a different kind of transformation for this. There would be two such symmetry transformations: each team plays towards the same goal, and each action starts on the right side of the pitch. Combining both could be particularly useful, for instance, when analyzing corners it is common to aggregate them at a common point, like (0,0).

JanVanHaaren:
Although I don't have much experience with orientations in `kloppy`, I like the proposal of @probberechts.

PR #400, opened 2025-01-14
Adding a transform function to overlay teams

douglaspampu:
Adding a new transformation function intended to position all players in a frame pointing to the same direction, overlaying them instead of the traditional orientation where tam A faces team B

This PR implements the issue https://github.com/PySport/kloppy/issues/253

Some questions I have about the issue and this implementation are:
- I am creating a new transform for the tracking dataset, would the events dataset benefit of this transformation as well?
- I am implementing this transformation using the flip_point function on the players facing left-to-right, does the flip_point does the symmetry transformation @probberechts suggested on the issue?
- From a functional point of view, if I overlay the players on the field this way, the teams would be overlaid, but when I have an attacker facing a defender the transform would position them in opposite places of the field (defender is moved to the right side of the pitch and attacker to the left side) is this the expected behavior?

Appreciate any comments on this @probberechts and @DriesDeprest


probberechts:
Let's illustrate how I think it should work. Assume we have a dataset with two corner kicks. One by the blue team and one by the red team. For each corner kick there is one defender in the penalty area.

Untitled-1-01

After a "MirrorHorizontal"[^1] operation, it should look like this.

Untitled-1-02

Next, after a "MirrorVertical" operation, it should look like this.

Untitled-1-03

> I am creating a new transform for the tracking dataset, would the events dataset benefit of this transformation as well?

Ideally, this works for both event and tracking data.

> I am implementing this transformation using the `flip_point` function on the players facing left-to-right, does the `flip_point` does the symmetry transformation @probberechts suggested on the issue?

Yes, you can use it for the "MirrorHorizontal" operation.


[^1]: I am not sure about the best name for this operation yet.

DriesDeprest:
To be clear, what I wanted to achieve with:  https://github.com/PySport/kloppy/issues/253 is the following:

image

douglaspampu:
Hi @probberechts and @DriesDeprest Thanks for your explanations and examples! this makes things a lot clearer.

One comment I would like to make is that each one of you have a different feature in mind. @probberechts is suggesting is applying horizontal and vertical transformations until we can overlay different actions, like the corner kicks in his example. For that, we could have the "mirror_vertical" and "mirror_horizontal" functions available for the user to apply over all players in a frame or event until they have these players rotated to a specific position.
On the corners, for example, we would apply mirror_vertical and mirror_horizontal on the top left corner. But we would apply just the mirror_horizontal on the top right corner.

On another hand, @DriesDeprest suggestion seems to be overlaying formations. I would rotate one of the teams for two players playing on the same position for opposed teams to be on the same position on the field. But, in this case, if I apply this transform for the corners I would rotate the corner kicker but not the defender. 
I can see the importance for this feature if I am analyzing just the attacking actions of one team over time regardless of the defenders of the other team (or even on different matches).

For this one, it would be possible to refactor the flip_point to call a "mirror_vertical" and "mirror_horizontal" functions and then also expose these new functions to allow for both suggestions

probberechts:
If I understand it correctly, what @DriesDeprest suggests is the combination of a MirrorHorizontal + MirrorVertical operation (in his example applied to the away team). I considered that both of these operations would be useful, hence why I initially proposed to split it up in those two transformations.

douglaspampu:
@probberechts and @DriesDeprest 
I refactored the code to do a horizontal mirror, flip point on the x axis, and a vertical mirror, flip over the y axis.

I kept the overlay option to flip the coordinates of the team facing left to right, but this is now using the the mirror operations.
I also exposed the mirror vertical and horizontal options on the transform function to allow the suggestion of rotating frames to have all the corners starting from the same position.

Appreciate any comment you have over this! If this implementation is good I will add support for events as well

PR #408, opened 2025-01-21
[HawkEye 2D] Support for HawkEye centroids and ball data

UnravelSports:
Hi all, 

I've used @probberechts [HawkEye Skeletal](https://github.com/probberechts/kloppy/tree/feat/hawkeye) branch and reworked it to simply support HawkEye 2D data. This means only the `football.samples.ball` and `football.samples.centroids` files and not the skeletal data.

Below is an outline of its behavior, since  it works slightly differently than other providers given the HawkEye file format.

### **Load Parameters**
`hawkeye.load()` has a couple of parameters:
- ball_feeds
- player_centroid_feeds
- meta_data = None
- pitch_width = 68.0
- pitch_length = 105.0
- sample_rate = None
- limit = None
- coordinates = None
- show_progress= False
 -------
### **ball_feeds** and **player_feeds**
This supports:
- Single file paths (e.g `ball_feeds="hawkeye_1_1.football.samples.ball"` and `player_feeds="hawkeye_1_1.football.samples.ball"`)
- List of file paths (e.g. `ball_feeds=["hawkeye_1_1.football.samples.ball", "hawkeye_2_46.football.samples.ball"`] etc.
- Folder path (e.g. "path/to/hawkeye/data")
It also supports any cloud based files or folder (bucket) locations as long as you have set up your credentials correctly. For this we rely on the `fsspec` package.

⚠️ Please note if you change the file names you're going to have a bad time, because for simplicity we sort the files by period and minutes in the file names as provided by HawkEye. 

-------
### **meta_data, pitch_width & pitch_length**
For some reason meta data is optional with HawkEye. This means that if we do have meta_data we can provide it here. We support `.xml` and `.json`. However, I have no idea how many different types of HawkEye meta data exist.

If we don't have meta data, or the meta data does not contain pitch dimensions we can provide it with `pitch_width` and `pitch_length`. Meta data pitch dimensions will overwrite passed parameters. (This seemed like the most logical choice). 

We further use the metadata only to grab `game_id`, `game_week` and `game_date` which we add to the kloppy Metadata.

*Note: FIFA Assumes standard pitch size, hence the default values of 105 and 68*

------- 

### **sample_rate**, **limit** and **coordinates**
These work as with any other Kloppy loader

--------
### **show_progress**
Uses tqdm to show loading progress. @probberechts added this because loading skeletal data would take a while, I didn't see a need to remove it.

-------

Additional changes I've made to Pieters' version:
- We use `self.__infer_frame_rate(ball_tracking_data)` to infer the frame_rate, since it's not provided anywhere (not even in the metadata)
- Because it seems HawkEye has some different possibilities for using Ids (for example "fifaId" or "uefaId") we use the `HawkEyeObjectIdentifier` as described below to take the ids in some sort of priority order. If fifaId is available we take that, if uefaId is available we take that and if neither exists we default to heId. At some point we might want to allow the user to pass their preferred Id directly.
```
class HawkEyeObjectIdentifier:
    FIFA_ID = "fifaId"
    UEFA_ID = "uefaId"
    HE_ID = "heId"

    PRIORITY_IDS = [FIFA_ID, UEFA_ID, HE_ID]

    @classmethod
    def get_identifier_variable(cls, player_tracking_data):
        player_object = player_tracking_data["details"]["players"][0]["id"]

        object_id = cls.HE_ID
        for identifier in cls.PRIORITY_IDS:
            if player_object.get(identifier):
                return identifier
        return object_id
```
- Timestamp is "normalized" to 0 at the start of every period.
- If have absolutely no idea what the use is of "segment: 0" when parsing the periods, so I'm simply skipping it to make `metadata.periods` be logical.

----

Edit: Woops forgot to merge master.


UnravelSports:
@koenvo @probberechts I'm getting a weird error in the Windows build environment related to Python3.12 and `fsspec`. I have no idea how to troubleshoot this since I don't have access to Windows.

UnravelSports:
I used `fsspec` @koenvo suggestion in the Slack channel.

koenvo:
The adapters don’t allow listing the adapters don’t allow listing of files. Could add it to the adapter but feels like reimplementing fsspec.

probberechts:
> The adapters don’t allow listing the adapters don’t allow listing of files. Could add it to the adapter but feels like reimplementing fsspec.

Yes, that was exactly what I was thinking. I'm also fine with using `fsspec` but do we need **both** the adapters and `ffspec`? I was also wondering whether it provides the same level of flexibility as the adapter pattern. For example, how can I set the password of a FTP server?

UnravelSports:
As far as the debate between fsspec and adapters, I have not enough knowledge on either to contribute anything of value.

As far as my failing tests go, it seems that `urlparse` inside `__init_fsspec()` parses the following path `ball_feed_file = WindowsPath('D:/a/kloppy/kloppy/kloppy/tests/files/hawkeye_1_1.football.samples.ball')` as being the "d" protocol (taken from `D:/`). I can't really find if this is a common issue (well, if I can't find it the answer is probably that it's not). So this might suggest there is something wrong with my implementation with `fsspec`. Again, I have no idea how this works, or how it supposed to work or all the things that it's supposed to be able to do or not. I just want to parse some football data into a dataframe 😆

probberechts:
@UnravelSports Ahah, no worries. I took a quick look at `fsspec`, and it seems like it can really simplify much of our I/O code. I'll refactor the code to use `fsspec` as the default for file handling. It should be straightforward (mostly deleting redundant code). I plan to have this done by the end of the week.

UnravelSports:
Pieter you are the best!

probberechts:
@UnravelSports I've added three failing tests. Could you check these?

UnravelSports:
@probberechts for the first test (setting the pitch dimensions by passing parameters) it seems you prefer to overwrite the pitch dimensions when they are passed to the deserializer even if we have meta data that contains pitch dimensions. I don't think this makes sense, and would prefer to have the meta data be leading. Simply because I don't foresee a scenario where we know better what the pitch dimensions are then the data provider. The `pitch_length` and `pitch_width` parameters are there because half the time HawkEye does not provide meta data and in these cases FIFA assumes 105 x 68. 

There should hardly be any reason to use these two parameters, but they are there because it's also strange to hard code the default 105x68 whenever no meta data is available.

This would mean that:

```python
dataset = hawkeye.load(
    ball_feeds=Path(ball_feed_file[0]),
    player_centroid_feeds=player_ball_file[0],
    pitch_length=120
)
```
Would set the pitch_length to 120 in the metadata, but also passing meta data should not use the 120.

```python
dataset = hawkeye.load(
    ball_feeds=Path(ball_feed_file[0]),
    player_centroid_feeds=player_ball_file[0],
    meta_data=meta_data_json,
    pitch_length=120
)
```
This should have pitch length in metadata at 104 (as available in the meta data).

probberechts:
When a user explicitly provides a parameter, they likely intend for it to override default behavior. It almost always works like this: parameters get precedence over things that are more "hidden" like general config settings, environment variables, etc. I don't think it's a good idea to deviate from this.

UnravelSports:
Hmm. I understand this is generally normal behavior (to give precedence), but it feels to me that in this case it simply doesn't make any sense whatsoever to want to deviate from the meta data provided pitch dimensions if they exist. Like, what is the point of setting the pitch to 120x70 if the meta data clearly says the pitch is actually 104x64? 

Tthe functionality for these two parameters is not to give the user the ability to overwrite existing metadata, but rather to give otherwise missing information if the meta data is not available.

Should we instead throw an error if two sets of pitch dimensions are provided? Stating that only 1 should be provided (aka the user should either remove the meta data or the pitch length/width parameters from their function call).

probberechts:
This is a very minor point to me. Personally, I do not see any harm in allowing users to override the metadata value to deal with incorrect values (I don't know whether that is a realistic problem). On the other hand, if you do not want to allow this, I am also ok with simply adding "ignored if metadata is provided" to the docstring of the pitch_widht and pitch_height parameters.

koenvo:
A solution could be to name it something like “hint”. This indicates the deserializer can use the value, but doesn’t have to.

Maybe another option is to infer it from the data? Although, this might break when using limit and sample rate etc

UnravelSports:
Inferring it from the data is a nightmare waiting to happen. Since it's a very minimal issue I simply went with @probberechts suggestion and I've updated the docstrings to read:

```
 pitch_length: The length of the pitch (in meters). Ignored if metadata that includes pitch length is provided.
 pitch_width: The width of the pitch (in meters). Ignored if metadata that includes pitch width is provided.
```

PR #411, opened 2025-01-23
[BUG] Incorrect MinutesPlayed aggregation

UnravelSports:
Hi,

Just noticed we were incorrectly assigning `_start_time` to `end_time` in the dataset aggregator, this lead to end_time and end_period being 0 and 1 respectively for all starting players, also those that got subbed off.

```python
MinutesPlayed(
      player=player,
      start_time=_start_time,
      end_time=end_time, #<- this was _start_time
      duration=end_time - _start_time,
  )
```

### DIFF
--- a/kloppy/domain/services/aggregators/minutes_played.py
+++ b/kloppy/domain/services/aggregators/minutes_played.py
@@ -49,7 +49,7 @@ class MinutesPlayedAggregator(EventDatasetAggregator):
                             MinutesPlayed(
                                 player=player,
                                 start_time=_start_time,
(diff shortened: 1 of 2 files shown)


UnravelSports:
@koenvo @probberechts I've added a simple test here. In the test we check fails if `duration.total_seconds()` is not equal to the difference between starting time and end time in seconds (we include exact period start / end times here). 

This fails in the original version, meaning `end_time=_start_time` is indeed wrong.

koenvo (approved):

PR #414, opened 2025-01-23
[BUG] Fix setting metadata after transform

UnravelSports:
As mentioned here #407 when calling `replace()` on the metadata the changes are not actually reflected in the new metadata object.

Aka, the below simply doesn't work:
```python
metadata = replace(
      dataset.metadata,
      coordinate_system=to_coordinate_system,
      pitch_dimensions=to_coordinate_system.pitch_dimensions,
      orientation=to_orientation,
  )
```
Since `Metadata` is not a `frozen` dataclass we can also update the existing metadata as follows:

```python
dataset.metadata.coordinate_system = to_coordinate_system
dataset.metadata.pitch_dimensions = to_pitch_dimensions
dataset.metadata.orientation = to_orientation
```

Finally, at the end of the transformation we do this:

```python
EventDataset(
      metadata=dataset.metadata, #<- This used to be the broken metadata object.
      records=events,
  )
```
-----

For some reason this messed up a test (see below) because "action_executing_team" was not set when calling `frame_t4.attacking_direction`. However, I think this is an issue with how the test was set up, because a TrackingDataset (which is used in this test) does not have the ability to be set to `ACTION_EXECUTING_TEAM`, and it should instead be transformed to `BALL_OWNING_TEAM` instead.

```
# Transform to ACTION_EXECUTING_TEAM orientation
# this should be identical to BALL_OWNING_TEAM for tracking data
transform4 = transform3.transform(
    to_orientation=Orientation.ACTION_EXECUTING_TEAM,
    to_pitch_dimensions=to_pitch_dimensions,
)
assert transform4.metadata.orientation == Orientation.ACTION_EXECUTING_TEAM
assert transform4.frames[1].ball_coordinates == Point3D(x=0, y=1, z=1)
        for frame_t3, frame_t4 in zip(transform3.frames, transform4.frames):
            assert frame_t3.ball_coordinates == frame_t4.ball_coordinates
            assert frame_t3.attacking_direction == frame_t4.attacking_direction
```

### DIFF
--- a/kloppy/domain/services/transformers/dataset.py
+++ b/kloppy/domain/services/transformers/dataset.py
@@ -400,6 +400,13 @@ class DatasetTransformer:
                 to_pitch_dimensions=to_pitch_dimensions,
                 to_orientation=to_orientation,
             )
+            if dataset.metadata.coordinate_system is not None:
+                dataset.metadata.coordinate_system.pitch_length = (
+                    to_pitch_dimensions.pitch_length
+                )
+                dataset.metadata.coordinate_system.pitch_width = (
+                    to_pitch_dimensions.pitch_width
+                )
             dataset.metadata.pitch_dimensions = to_pitch_dimensions
             dataset.metadata.orientation = to_orientation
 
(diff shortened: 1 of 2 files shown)


UnravelSports:
Note: This seems to still not work 100% correctly, because for some reason we store pitch dimensions in two places, namely:

- `dataset.metadata.pitch_dimensions`
- `dataset.metadata.coordinate_system`

The former is fixed, but using `dataset.metadata.coordinate_system` still gives the old pitch dimensions

Issue #413, opened 2025-01-23
Add distance/length and angle for events with end coordinates


lode-mgp:
Would it be a good idea to add properties that can be derived from the start and end coordinates of an event?

I am thinking about two properties:
- distance (between start and end coordinates)
- angle to goal

There could be multiple conventions/orientations for the angle but I think the following would be a good default.

![Image](https://github.com/user-attachments/assets/4c48355c-f299-4b7a-a2c8-6aaf155b8a20)

Do you think it makes sense to add these properties in kloppy or not because these properties do not really add extra information since they can be derived from the already existing properties?

UnravelSports:
They are not very well documented I think, but they should already exist.

For example, as per [this test](https://github.com/PySport/kloppy/blob/master/kloppy/tests/test_to_records.py), you should be able to do:

```
from kloppy.domain.services.transformers.attribute import (
    DistanceToGoalTransformer,
    DistanceToOwnGoalTransformer,
    AngleToGoalTransformer
)

dataset.filter("pass").to_records(
        "timestamp",
        "player_id",
        "coordinates_*",
        DistanceToGoalTransformer(),
        DistanceToOwnGoalTransformer(),
        AngleToGoalTransformer()
    )
```

lodevt:
Hi,

Thanks for clarifying this! There seems to be a bug in the implementation of the `AngleToGoalTransformer`. I think [this line](https://github.com/PySport/kloppy/blob/19faf69f9fa611e9739724368337f111a214bcb8/kloppy/domain/services/transformers/attribute.py#L60) (and other lines beneath) should reference 'pitch_width' and 'pitch_length' instead of just 'width' and 'length', as these attributes don't exist.

Also, the current implementation only allows to calculate Angle when the orientation is 'ACTION_EXECUTING_TEAM'.

I could create a PR to
 - fix the bug 
 - add support to calculate Angle for other orientations 
 - add a new transformer to calculate the distance covered by an event (if start and end coordinates are present) as this is currently not available (only distance to goal/own goal are)

UnravelSports:
I would not be surprised if this indeed is broken, because I think this is _very_ old functionality. @koenvo would you mind chiming in on this?

I think having this functionality, as well as the delta x,y values makes sense would you agree? 

Also @probberechts

probberechts:
Yes, it makes sense to fix any bugs and add a transformer to compute the distance.

[...]

PR #416, opened 2025-01-24
[BUG] - use correct attributes in AngleToGoalTransformer

lodevt:
Fixing the bug mentioned in [Issue 413](https://github.com/PySport/kloppy/issues/413).

### DIFF
--- a/kloppy/domain/services/transformers/attribute.py
+++ b/kloppy/domain/services/transformers/attribute.py
@@ -57,12 +57,16 @@ class AngleToGoalTransformer(EventAttributeTransformer):
         if not event.coordinates:
             return {"angle_to_goal": None}
 
-        if metadata.pitch_dimensions.width:
+        if metadata.pitch_dimensions.pitch_width:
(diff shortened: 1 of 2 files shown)


probberechts:
This is still not correct. The `pitch_length` and `pitch_width` are independent of the dimensions that are used for the coordinate system.  For example, the `PitchDimensions` can have `pitch_length=105, pitch_width=68` but the coordinates could still be in the range `[0,1]`. You have to look at the `PitchDimensions.standardized` field.

lode-mgp:
It should now be consistent for all coordinate systems by converting the delta_x and delta_y to meters before calculating the angle, using the pitch_dimensions from the metadata.

lodevt:
It should now be consistent for all coordinate systems by converting the delta_x and delta_y to meters before calculating the angle, using the pitch_dimensions from the metadata. The angle is always a value between 0 and 90 degrees.

probberechts:
Thanks!

PR #419, opened 2025-01-25
Add `io.expand_inputs`; refactor adapters to use `fsspec`

probberechts:
- Refactors the `kloppy.infra.io.adapters` to use [fsspec](https://filesystem-spec.readthedocs.io/en/latest/). This simplifies the code and makes it easier to add other filesystems (that are supported by ffspec).
- Adds an adapter for zip archives
- Adds a `kloppy.io.expand_inputs` method that maps a directory, list of paths, ... to a list of file-like objects that can be opened with `kloppy.io.open_as_file`. This should be useful for #408 and #356 

```python
>>> from kloppy import io
>>> list(io.expand_inputs("data"))
['/cw/dtaijupiter/NoCsBack/dtai/pieterr/Projects/kloppy-dev/data/period1_min1.txt',
 '/cw/dtaijupiter/NoCsBack/dtai/pieterr/Projects/kloppy-dev/data/period1_min2.txt',
  ...
 '/cw/dtaijupiter/NoCsBack/dtai/pieterr/Projects/kloppy-dev/data/period2_min10.txt'
]
>>> list(io.expand_inputs(["data"]))
['/cw/dtaijupiter/NoCsBack/dtai/pieterr/Projects/kloppy-dev/data/period1_min1.txt',
 '/cw/dtaijupiter/NoCsBack/dtai/pieterr/Projects/kloppy-dev/data/period1_min2.txt',
  ...
 '/cw/dtaijupiter/NoCsBack/dtai/pieterr/Projects/kloppy-dev/data/period2_min10.txt'
]
>>> list(io.expand_inputs(["data/period1_min1.txt"]))
['data/period1_min1.txt']
>>> list(io.expand_inputs("data", regex_filter="period2"))
['/cw/dtaijupiter/NoCsBack/dtai/pieterr/Projects/kloppy-dev/data/period2_min1.txt',
 '/cw/dtaijupiter/NoCsBack/dtai/pieterr/Projects/kloppy-dev/data/period2_min2.txt',
  ...
 '/cw/dtaijupiter/NoCsBack/dtai/pieterr/Projects/kloppy-dev/data/period2_min10.txt']
>>> list(io.expand_inputs("data", regex_filter="period2", sort_key=lambda x: x))
['/cw/dtaijupiter/NoCsBack/dtai/pieterr/Projects/kloppy-dev/data/period2_min1.txt',
 '/cw/dtaijupiter/NoCsBack/dtai/pieterr/Projects/kloppy-dev/data/period2_min10.txt',
 ...
 '/cw/dtaijupiter/NoCsBack/dtai/pieterr/Projects/kloppy-dev/data/period2_min9.txt']
>>> list(io.expand_inputs([open("data/period1_min1.txt")]))
[<_io.TextIOWrapper name='data/period1_min1.txt' mode='r' encoding='UTF-8'>]
```


kloppy/infra/serializers/tracking/skillcorner.py
```diff
@@ -237,17 +237,24 @@ def __replace_timestamp(obj):
...
+            data = json.load(file)
+            for line in data:
+                line = self.__replace_timestamp(line)
+            return data
+        except json.JSONDecodeError:
+            pass  # It's not a standard JSON document
+
+        try:
             data = []
             for line in file:
```
koenvo, line 250: This will probably break. The `json.load` will consume the file, and the `for line in data` won't be able to read anything. The `file` should support a rewind, and `file.seek(0)` needs to be added.

UnravelSports: This implementation of the SkillCorner file type checking and replacing the timestamp key (by me...) is not optimal in general. I can create a PR to do this the way we do it in other Deserializers (like tracab meta data xml / json checking) and then I'll figure out a nicer way to handle the time / timestamp key issue a bit more elegantly too.

probberechts: Indeed. Good catch! This should be fixed now. It might also be good to add a test cases for the JSONL format but unfortunately I do not have any recent SkillCorner data.

probberechts:
> I'm curious about the decision to implement caching specifically in S3Adapter and HTTPAdapter rather than only in FSSpecAdapter, which would apply it to all subclasses. What was the rationale behind this approach?

Caching is only applicable to remote file systems, which means implementing it directly in the `FSSpecAdapter` wouldn't have been feasible. I could have introduced an additional layer, such as a `RemoteFSSpecAdapter`, but I don't really see immediately how to implement that elegantly or how it could really simplify the current implementation.

koenvo:
> > I'm curious about the decision to implement caching specifically in S3Adapter and HTTPAdapter rather than only in FSSpecAdapter, which would apply it to all subclasses. What was the rationale behind this approach?
> 
> Caching is only applicable to remote file systems, which means implementing it directly in the `FSSpecAdapter` wouldn't have been feasible. I could have introduced an additional layer, such as a `RemoteFSSpecAdapter`, but I don't really see immediately how to implement that elegantly or how it could really simplify the current implementation.

Totally agree. More layers makes it more complex. 

PR looks good te me. Thanks!

Issue #415, opened 2025-01-23
Add support for inserting events in a `EventDataset`


probberechts:
A recurring requirement seems to be having the ability to insert events into an existing `EventDataset`. Some use cases are:

- Inserting carries generated from other events in the dataset. (#396) 
- Inserting substitutions with approximate timestamps (#368).
- Inserting pressing events derived from a tracking dataset.

To support these needs, I propose adding an `insert` method to the EventDataset interface. 

Sometimes the correct position can be easily determined by parameters such as the index, the ID of the preceding or succeeding event. or an exact timestamp. However, when the exact position is unknown, the `insert `method should be able to determine the optimal insert location based on a set of constraints. For example, if the approximate time of a substitution is known, the event should be inserted within a window where the ball state is dead, closest to the given timestamp. Similarly, pressing events should be inserted into sequences where the opponent team is in possession.

----

Below is the API that Copilot suggests. I'm not sure whether this is exactly what we want, but I think it provides a good starting point.

```py
class EventDataset:
    def insert(
        self, 
        event: Event, 
        position: Optional[int] = None, 
        before_event_id: Optional[str] = None, 
        after_event_id: Optional[str] = None, 
        timestamp: Optional[datetime] = None, 
        constraints: Optional[Callable[[Event, EventDataset], bool]] = None
    ) -> None:
        """
        Inserts an event into the dataset at the appropriate position.

        Parameters:
        ----------
        event : Event
            The event to be inserted into the dataset.
        
        position : Optional[int], default=None
            The exact index where the event should be inserted. If provided, 
            overrides all other positioning parameters.
        
        before_event_id : Optional[str], default=None
            The ID of the event before which the new event should be inserted. 
            Ignored if `position` is provided.

        after_event_id : Optional[str], default=None
            The ID of the event after which the new event should be inserted. 
            Ignored if `position` or `before_event_id` is provided.

        timestamp : Optional[datetime], default=None
            The timestamp of the event, used to determine its position based 
            on chronological order if no other positional parameters are specified.

        constraints : Optional[Callable[[Event, EventDataset], bool]], default=None
            A custom function that takes the event and dataset as arguments and 
            evaluates whether the event satisfies specific conditions to determine 
            its position. Useful for more complex insertion logic, such as inserting 
            into a valid contextual window (e.g., dead ball states or possession sequences).

        Returns:
        -------
        None
            The method modifies the dataset in place.

        Raises:
        ------
        ValueError
[...]

PR #421, opened 2025-01-29
Add support for inserting events in a EventDataset

lodevt:
An initial effort to implement the functionality from [Issue 415](https://github.com/PySport/kloppy/issues/415).

To insert events based on a set of constraint, I opted to use a scoring function. The scoring function assigns a value to each existing event that indicates how suitable that position is for the new event to be inserted (higher is more suitable).

So for example, when you want to insert a pressing event into a sequence where the opponent team is in possession, as close as possible to a given timestamp, your scoring function could look like this
```python
def scoring_function(event: Event, dataset: EventDataset):
    if event.state["sequence"].team == event_to_insert.team:
        return 0
    if event.period != event_to_insert.period:
        return 0
    return 1 / abs(
        event.timestamp.total_seconds()
        - event_to_insert.timestamp.total_seconds()
    )
```


probberechts:
This is a very nice solution!

I only had to add some code to update the pointers to the previous / next event. My implementation is a bit of a hack, though. I would rather use the [`set_refs`](https://github.com/PySport/kloppy/blob/aa74447789e14a74fc8a4959ab0d4be99d8c3956/kloppy/domain/models/common.py#L1079) method, but I didn't immediately see how to make it work for both the `filter` (you probably don't want to update the refs) and the `insert` (you want to update the refs) methods. Does anyone have a a better solution?

One other thing I just realized now is that it will be unclear to the user whether the new event will be inserted before or after the event that gives the max score. I guess it will be sufficient to specify this in the docstring.

lodevt:
I also don't immediately see a very clean solution. One alternative could be adding a `force_update` parameter to the `set_refs` method, setting it to `True`in this case. I don't think this is necessarily better than your implementation, but it offers an alternative and allows to reuse the `set_refs`method.

````python
    def set_refs(
        self,
        dataset: "Dataset",
        prev: Optional["DataRecord"],
        next_: Optional["DataRecord"],
        force_update: bool = False,
    ):
        if hasattr(self, "dataset") and not force_update:
            # TODO: determine if next/prev record should be affected
            # by Dataset.filter
            return

        self.dataset = dataset
        self.prev_record = prev
        self.next_record = next_

probberechts (approved):

koenvo:
I will check the PR in more detail, but was already thinking about how this works.

from a workflow perspective I believe the insert should happen right after loading, and generates an  better/improved/corrected truth of what happened during the game.

After the dataset is corrected people can start processing like filter, transform etc. Small note: we should put a flag on a dataset to indicate it’s filtered?

From that perspective I believe correcting actions, like inserts, should update the refs. The original was incorrect and so are the refs. 

Curious to your opinions!

probberechts:
Yes, good point: it's generally not a good idea to insert events into a filtered dataset. Adding a flag is a great solution!

However, there may be cases where transforming the data first makes sense before inserting events. For example, if you're manually creating event annotations using a specific coordinate system, you’d want to transform the dataset to that coordinate system before adding the events to ensure consistency.

On a related note, the coordinates of inserted events must be consistent with the dataset’s coordinate system. I guess there is currently no way to easily verify or enforce this?

PR #427, opened 2025-02-14
[Wyscout V3] Add shot result z coordinates

DriesDeprest:
I extended the parsing of Wyscout V3 shot result coordinates to also contain z coordinates. I also now recognize shots on the posts and give them the relevant result coordinates.

### DIFF
--- a/kloppy/infra/serializers/event/wyscout/deserializer_v3.py
+++ b/kloppy/infra/serializers/event/wyscout/deserializer_v3.py
@@ -4,7 +4,7 @@ import warnings
 from dataclasses import replace
 from datetime import datetime, timedelta, timezone
 from enum import Enum
-from typing import Dict, List, Optional
+from typing import Dict, List, Optional, Union
 
 from kloppy.domain import (
     BodyPart,
@@ -39,6 +39,7 @@ from kloppy.domain import (
     CarryResult,
     ExpectedGoals,
     PostShotExpectedGoals,
+    Point3D,
 )
 from kloppy.exceptions import DeserializationError, DeserializationWarning
 from kloppy.utils import performance_logging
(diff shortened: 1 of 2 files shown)


kloppy/infra/serializers/event/wyscout/deserializer_v3.py
```diff
@@ -110,24 +111,74 @@ def _flip_point(point: Point) -> Point:
...
+        """Returns the string identifier of the shot result."""
+        return self._code
+
+    @property
+    def point(self) -> Point3D:
+        """Returns the predefined Point3D associated with the shot result."""
+        return self._point
+
+    @classmethod
+    def point_from_code(cls, code: str) -> Union[Point3D, None]:
```
probberechts, line 175:
```suggestion
    def point_from_code(cls, code: str) -> Optional[Point3D]:
```

kloppy/infra/serializers/event/wyscout/deserializer_v3.py
```diff
@@ -110,24 +111,74 @@ def _flip_point(point: Point) -> Point:
...
+
+    @property
+    def code(self) -> str:
+        """Returns the string identifier of the shot result."""
+        return self._code
+
+    @property
+    def point(self) -> Point3D:
+        """Returns the predefined Point3D associated with the shot result."""
+        return self._point
```
probberechts, line 172: Minor, but I think using properties here doesn't have any added value.
 ```suggestion
        self.code = code
        self.point = point
```

File: kloppy/infra/serializers/event/wyscout/deserializer_v3.py [hunk 3]
```diff
@@ -110,24 +111,74 @@ def _flip_point(point: Point) -> Point:
...
+     -------------------------------
+          || glt | gt | grt ||            2
+     --------------------------------
+      ol || gl | gc  | gr || or           1
+     --------------------------------
+      olb || glb  | gb | grb || orb       0
+
+      40     45    50    55     60    (y-coordinate of zone)
+        44.62               55.38     (y-coordinate of post)
+    """
```
probberechts, line 133: You can document the attributes here:
```suggestion
   Attributes:
          code (str): The string identifier of the shot result.
          point (Point3D): The predefined Point3D associated with the shot result.
    """
```

kloppy/infra/serializers/event/wyscout/deserializer_v3.py
```diff
@@ -110,24 +111,74 @@ def _flip_point(point: Point) -> Point:
...
+    def point(self) -> Point3D:
+        """Returns the predefined Point3D associated with the shot result."""
+        return self._point
+
+    @classmethod
+    def point_from_code(cls, code: str) -> Union[Point3D, None]:
+        """Retrieves the enum member from a string code."""
+        for zone in cls:
+            if zone.code == code:
+                return zone.point
```
probberechts, line 179: I think you can avoid the loop here:
```suggestion
        if code in cls.__members__:
            return cls[code].point
        warnings.warn(f"Invalid shot zone code: {code}")
        return None
```

DriesDeprest:
```python
        if code in cls.__members__:
            return cls[code].point
        warnings.warn(f"Invalid shot zone code: {code}")
        return None
```
didn't work for me, so I kept it as is.

PR #431, opened 2025-02-18
[BUG] Disassociate sample_rate and limit 

UnravelSports:
Arguably the most minor of PRs, but I noticed that some tracking deserializers (e.g. Sportec) would return n+1 records when setting limit as n.

Additionally, it annoyed me that the sample rate would impact the limit. This meant that a sample_rate of (1/2) would give you only 50 samples if you set the limit to 100.

I've disassociated the sample_rate from the limit. This means if we set the limit to 100 and the sample rate to (1/2) we still return 100 records, but they are still sampled at a rate of one in two.

I've added tests to all tracking providers for which this was handled incorrectly.


kloppy/tests/test_statsperform.py
```diff
@@ -277,6 +277,32 @@ def test_framerate(self, tracking_dataset: TrackingDataset):
     def test_flags(self, tracking_dataset):
         assert tracking_dataset.metadata.flags == DatasetFlag.BALL_STATE
 
+    def test_coordinate_system_with_pitch_dimensions(
+        self, tracking_data: Path, tracking_metadata_xml: Path
+    ):
```
probberechts, line 282: The test name doesn't describe what the test is doing here.

kloppy/tests/test_metrica_epts.py
```diff
@@ -98,6 +98,24 @@ def meta_data(self, base_dir) -> str:
...
+            sample_rate=(1 / 2),
+        )
+        assert len(dataset.records) == 50
+
+        dataset = metrica.load_tracking_epts(
+            meta_data=meta_data,
+            raw_data=raw_data,
+            limit=100,
+        )
+        assert len(dataset.records) == 100
```
probberechts, line 117: In all these test cases, the order isn't very logical. I would be more logical to first test the "easy" scenario where only a `limit` is provided.

UnravelSports:
I have updated the ordering in all tests, fix the test name, and updated the way we handle statsperform frame counting (no more `enumerate`, but instead we use `n_frames` like in all other tracking data deserializers)

UnravelSports:
And I merged master to fixed the broken PFF test 🤞

Issue #362, opened 2024-11-07
Substitutions, players temporarily leaving pitch in Kloppy EventData model and dataframe


ussf-jbekkers:
Before 3.16.0, in the Opta Parser specifically, `PLAYER_ON` and `PLAYER_OFF` are used to describe a substitution with these two individual events.

In (currently officially not yet released) version [3.16.0](https://github.com/PySport/kloppy/milestones) we add a breaking change to this by adding the `SUBSTITUTION` event. Where the `SUBSTITUTION` event is now the combination of `PLAYER_ON` and `PLAYER_OFF` events in case of a substitution. However, the `PLAYER_ON` and `PLAYER_OFF` events can also relate to players temporarily retiring from the pitch as shared by @DriesDeprest in #361.

Some ideas for behavior as discussed by @DriesDeprest @probberechts and me for the Kloppy EventData model should be:
- Make `PLAYER_ON` and `PLAYER_OFF` more specific to mean `PLAYER_RETIRES` and `PLAYER_RETURNS` for any event that is unrelated to a substitution.
- In case of a substitution use the `SUBSTITUTION` event and assign the `replacement_player` the Id of the player coming on. (This is implemented in a commit in [#333](https://github.com/PySport/kloppy/pull/333/commits/003e5567fdd84e3f2af672912a15b95c1cecf71b) 
- Deprecate the use of `PLAYER_ON` and `PLAYER_OFF` explicitly as they are confusing and ambiguous. Similarly to how we did away with `HOME_AWAY` orientation in favor of `STATIC_HOME_AWAY` and they were explicitly renamed.

Idea for the implementation in the dataframe representation:
- Set the `replacement_player` in the `receiving_player` field for the `SUBSTITUTION` event.
- Remove the `GENERIC:player off` and `GENERIC:player on` in favor of `GENERIC:player retires` and `GENERIC: player returns` events respectively.

Let me know if I forgot anything or if anyone has any suggestions on this!

probberechts:
There are two general types of events that are related to players going off the pitch and returning on the pitch. 

1. First, there are substitutions. Substitutions are final. A player cannot return after he has been substituted. 
2. Second, there are various reasons for which a player temporarily has to leave the pitch. Typically, these are related to injury treatments. 

Hence, I would argue that we need at least two different event types.

**Substitutions**
[...]

PR #440, opened 2025-03-17
[Bugfix] Support substitution events with no replacement player

DriesDeprest:
Currently, in our StatsPerform deserializer we potentially set the replacement_player as None in case the player off event is not directly followed by a player on event, which happens in approx 1% of the matches, see: https://github.com/PySport/kloppy/blob/master/kloppy/infra/serializers/event/statsperform/deserializer.py#L396-L406. 

In order to support substitution events where no new sub on player is coming on the pitch for the sub off player, we add a check to see whether the replacement player actually exists prior to updating its position.

For context, a related discussion on substitution events: https://github.com/PySport/kloppy/issues/362.

### DIFF
--- a/kloppy/domain/models/event.py
+++ b/kloppy/domain/models/event.py
@@ -1076,18 +1076,22 @@ class EventDataset(Dataset[Event]):
 
         for event in self.events:
             if isinstance(event, SubstitutionEvent):
-                if event.replacement_player.starting_position:
-                    replacement_player_position = (
-                        event.replacement_player.starting_position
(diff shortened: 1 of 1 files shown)


probberechts:
If there is no other player coming on the pitch, it should not be counted as a subsitution, right? In Stats Perform data, these should be "Player retired" events.

Do you know why the `replacement_player` is missing? I have two hypotheses:

1. There is a player off event, but not necessarily as the next event? This should be addressed in the deserializer.
2. The event data is incomplete

Issue #429, opened 2025-02-15
Drop Python 3.8 support; Add Python 3.13 support


probberechts:
- Python 3.8 has reached it's end-of-life in October 2024. I propose to drop support for 3.8 in the next release and remove the tests. 
- People are starting to adopt Python 3.13. We should update the CI/CD pipelines to test against Python 3.13.

PR #447, opened 2025-03-31
[SPORTEC] Fixed Event Coordinate vertical_orientation (new)

UnravelSports:
Hi all, 

While writing the new docs @razor3598 noticed that in `SportecEventDataCoordinateSystem` the `vertical_orientation` was set to `VerticalOrientation.TOP_TO_BOTTOM`

```
@property
    def vertical_orientation(self) -> VerticalOrientation:
        return VerticalOrientation.TOP_TO_BOTTOM
```

We didn't know the correct answer. However, while I was working on the docs too when syncing events to tracking data I got this.

The ❌ here marks the event, and the jersey number indicates the event player in the tracking data.

### Upside Down
upside-down-3
upside-down-2
upside-down

This has to mean that the Events are oriented in the wrong vertical orientation.

### Correct

correct

FWIW, the tracking data is oriented correctly, which I checked on multiple frames while referencing player positioning.

### DIFF
--- a/kloppy/domain/models/common.py
+++ b/kloppy/domain/models/common.py
@@ -698,7 +698,7 @@ class SportecEventDataCoordinateSystem(ProviderCoordinateSystem):
 
     @property
     def vertical_orientation(self) -> VerticalOrientation:
-        return VerticalOrientation.TOP_TO_BOTTOM
+        return VerticalOrientation.BOTTOM_TO_TOP
 
     @property
     def pitch_dimensions(self) -> PitchDimensions:


UnravelSports:
It looks like a test in the PR is failing because of Python 3.8 and fsspec 
```
/opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/site-packages/fsspec/caching.py:40: in <module>     MultiFetcher = Callable[list[[int, int]], bytes]  # Maps [(start, end)] to bytes E   TypeError: 'type' object is not subscriptable
```
[StackOverflow](https://stackoverflow.com/questions/75202610/typeerror-type-object-is-not-subscriptable-python)

probberechts:
See [#1816](https://github.com/fsspec/filesystem_spec/issues/1816). We could set a constraint in the setup file but I suggest to simply drop Python 3.8 support (#429).

probberechts:
@UnravelSports If you merge with master, the tests should pass.

PR #460, opened 2025-05-14
[Impect] add support for Impect as event provider

DriesDeprest:
Adding support for Impect as an event provider. I've used the implementation of the StatsBomb deserializer as an inspiration source.

Happy to get feedback on the implementation.


DriesDeprest:
Closing till further notice.

PR #468, opened 2025-05-26
[SPORTEC] Fix missing `players_data` 

UnravelSports:
As discussed on Slack PR https://github.com/PySport/kloppy/pull/463 introduced a bug into the parser because players don't have a BALL_STATUS attribute.

Because we cannot assume Ball is the first object we see for a given frame we first create a dictionary with ball status per frame id. It's a bit of a silly solution, not sure if there is a better fix.

After we have this we reset the raw_data and do the full iteration again with the correct ball status.

I've added a test to make sure the players_data is not empty (this fails for the old implementation).

### DIFF
--- a/kloppy/infra/serializers/tracking/sportec/deserializer.py
+++ b/kloppy/infra/serializers/tracking/sportec/deserializer.py
@@ -95,15 +95,47 @@ def _unstack_framesets(
         This function assumes that the framesets are ordered by period and
         that the frames in each frameset are ordered.
     """
+    # First pass: collect ball status for each frame
+    ball_status_by_frame = dict()
+    context = etree.iterparse(
+        raw_data, events=("start", "end"), huge_tree=True
+    )
+
+    for event, elem in context:
+        try:
+            if event == "start" and elem.tag == "FrameSet":
+                current_obj = (
+                    "ball"
+                    if elem.get("TeamId") == "BALL"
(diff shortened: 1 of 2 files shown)


probberechts:
The extra pass made the code more complicated and slower:

```py
>>> from kloppy import sportec

# note: the data is already cached

# The version in this PR, loading only "alive" data
>>> %time tracking_dataset = sportec.load_open_tracking_data(match_id="J03WN1", only_alive=True, coordinates="sportec")
CPU times: user 38.8 s, sys: 1.06 s, total: 39.9 s
Wall time: 39.9 s

>>> tracking_dataset.to_df().shape
(68377, 125)

# The original code, loading all data
>>> %time tracking_dataset = sportec.load_open_tracking_data(match_id="J03WN1", only_alive=True, coordinates="sportec")
CPU times: user 14.7 s, sys: 324 ms, total: 15 s
Wall time: 15 s

>>>  tracking_dataset.to_df().shape
(141561, 125)
```

I think the most optimal and easiest thing to do is simply remove the ` and (not only_alive or int(elem.get(BALL_STATUS, 0)) == 1)` check from `_unstack_framesets`. Frames in which the ball is not in play are then filtered out while constructing the `Frame` entities.

UnravelSports:
If I do that the tests fail because we get 99 instead of 100 frames, any idea why this is?

PR #477, opened 2025-06-02
[SIGNALITY] Signality period times as `deltatime`

UnravelSports:
Just noticed in the recent Signality addition the Period times are in UTC instead of timedelta. 

To standardize this should be changed. I have also updated `Period` to such that `start_timestamp` and `end_timestamp` no longer have two different time options (removed `datetime`). 

I'm working on a different PR that re-introduces date times as `start_timestamp_utc` or something, because after an update all this information was lost. (It was still mentioned in the docstrings). 

Have these actual timestamps for periods should make it trivial to recompute the actual timestamps for each frame or event by simply adding timedelta to utc time, but currently we don't have this ability.

### DIFF
--- a/kloppy/domain/models/time.py
+++ b/kloppy/domain/models/time.py
@@ -24,22 +24,19 @@ class Period:
     Attributes:
         id: `1` for first half, `2` for second half, `3` for first half of
             overtime, `4` for second half of overtime, `5` for penalty shootout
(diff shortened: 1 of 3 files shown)


UnravelSports:
I think you are right, not sure why, but when I was "fixing" this it made sense, but now it totally doesn't.

Issue #476, opened 2025-06-02
Measure minutes played in and out of possession


lodevt:
The current `MinutesPlayedAggregator` [(see here)](https://github.com/PySport/kloppy/blob/master/kloppy/domain/services/aggregators/minutes_played.py) calculates a general `MinutesPlayed` and optionally also a `MinutesPlayedPerPosition`. I think it would make sense to extend this to also calculate `MinutesPlayerPerPossessionState`, which would provide minutes played for three possession states:
- in possession
- out of possession 
- ball dead

The existing ball_state and ball_owning_team properties can be used to extract the necessary information.

Example data structure: 
````python
class PossessionState(Enum): 
    IN_POSSESSION = 'in-possession'
    OUT_OF_POSSESSION = 'out-of-possession'
    BALL_DEAD = 'ball-dead'

class MinutesPlayedPerPossessionState(NamedTuple):
    player: Player
    possession_state: PossessionState
    duration: timedelta

````

koenvo:
Absolutely love this idea!

Would it be possible to make this more generic? Can imagine you also want to do this on team level. Or maybe only on possession state level, to check real playing time.

lodevt:
> Absolutely love this idea!
> 
> Would it be possible to make this more generic? Can imagine you also want to do this on team level. Or maybe only on possession state level, to check real playing time.

I think we could make it more generic by defining a general `executor` that can represent either a `Player` or a `Team`. The same data structure can be reused for both executors (of course the `MinutesPlayedPerPosition` would only be calculated for players). 


````python
class MinutesPlayed(NamedTuple):
    executor: Union[Player, Team]
    start_time: Time
    end_time: Time
    duration: timedelta
````

koenvo:
What if we use `key` instead of `executor` and allow a `tuple` / `NamedTuple` as key?

lode-mgp:
To make sure we are alligned, do you suggest using a class like this as `key` (replacing the `player`field)?

````python
class MinutesPlayedKey(NamedTuple):
    player: Optional[Player] = None
    team: Optional[Team] = None

class MinutesPlayed(NamedTuple):
    key: MinutesPlayedKey
    start_time: Time
    end_time: Time
    duration: timedelta
````

lode-mgp:
Can you give an example of how this might break typechecking?

koenvo:
[...]

PR #479, opened 2025-06-05
Extend MinutesPlayedAggregator with minutes played per possession state

lode-mgp:
This PR represents an effort to implement the functionality described in [Issue 476](https://github.com/PySport/kloppy/issues/476).

- I refactored the MinutesPlayedAggregator to make it more flexible and extensible. 
- I added an option to calculate minutes played per possession state (in possession, out of possession and dead ball).

Some notes:
- The effective play time per game falls in the range of 55-65 which seems to be realistic.
- This PR involves a breaking change, as the `include_position` flag is no longer supported and `player` inside `MinutesPlayed` is replaced by a more generic `key`. For example, this breaks the [example in the documentation](https://kloppy.pysport.org/examples/aggregations/#extract-minutes-played), since you can no longer access `item.player` but should use `item.key.player`. How should we handle this?

### DIFF
--- a/kloppy/domain/models/event.py
+++ b/kloppy/domain/models/event.py
@@ -978,13 +978,13 @@ class PassEvent(
         event_type (EventType): `EventType.PASS`
         event_name (str): `"pass"`
         result (PassResult): The pass's outcome.
-        receive_timestamp (Time): The time the pass was received.
+        receive_timestamp (timedelta): The time the pass was received.
         receiver_coordinates (Point): The coordinates where the pass was received.
         receiver_player (Player): The intended receiver of the pass.
         qualifiers: A list of qualifiers providing additional information about the pass.
     """
 
-    receive_timestamp: Optional[Time] = None
+    receive_timestamp: Optional[timedelta] = None
     receiver_player: Optional[Player] = None
     receiver_coordinates: Optional[Point] = None
 
@@ -1036,13 +1036,13 @@ class CarryEvent(
     Attributes:
         event_type (EventType): `EventType.CARRY`
         event_name (str): `"carry"`
-        end_timestamp (Time): Duration of the carry.
+        end_timestamp (timedelta): Duration of the carry.
         end_coordinates (Point): Coordinate on the pitch where the carry ended.
         result (CarryResult): The outcome of the carry.
         qualifiers: A list of qualifiers providing additional information about the carry.
     """
 
-    end_timestamp: Time
+    end_timestamp: timedelta
     end_coordinates: Point
 
     @property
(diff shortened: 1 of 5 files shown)


lodevt:
Thanks for your feedback @UnravelSports!
> I haven't fully digested this, but shouldn't the following also be in `RESULTS_CAUSING_DEAD_BALL` or `RESULTS_CAUSING_DEAD_BALL`:
> 
> * `PassResult.OUT`
> * `ShotResult.OFF_TARGET`
> * `InterceptionResult.OUT`
> * `EventType.BALL_OUT` (not sure why this exists next to `PassResult.OUT` and `InterceptionResult.OUT`, but that's beside the point.
> * `SetPieceType` in case providers mistag or miss an event prior?
> 

The results/event types you mentioned indeed also cause a dead ball, but strictly speaking, we don't need to include them for the results to be correct because of the following:
  - the ball state of`EventType.BALL_OUT` already is `BallState.DEAD` (so adding it it in the `EVENTS_CAUSING_DEAD_BALL`wouldn't make a difference)
  - events with `PassResult.OUT`, `ShotResult.OFF_TARGET` or `InterceptionResult.OUT` are always followed by a `BALL_OUT` event 
  
  So if the provider correctly tags "OUT" events, then it is not necessary to include these results in the `RESULTS_CAUSING_DEAD_BALL`.
  
  > I say the above with the notion that some data providers are really bad at tagging "OUT" events. 
> 

But of course, if the `BALL_OUT` events are often missing, then this makes it indeed necessary to include more items in the `EVENTS_CAUSING_DEAD_BALL` / `RESULTS_CAUSING_DEAD_BALL`.
  
> Additionally (and I think the answer is no), but is there a way to account for the fact that the `Result` has the time stamp of the action and not the "end timestamp" / exact moment it happened? Perhaps this works for `EventType.BALL_OUT` but not for `PassResult.OUT` since it's an outcome associated with a Pass that happened prior.
> 

 By excluding `PassResult.OUT` from `RESULTS_CAUSING_DEAD_BALL` (in the assumption this is always followed by `BALL_OUT` event), this problem would be solved (in the assumption each data provider correctly tags out events). However, it would still exist in other cases (for example `PassResult.OFFSIDE`). I think we can account for this by using the `receive_timestamp`(if this information is provided by the data provider). But if the `receive_timestamp` is missing, I also don't really see a way to account for this inaccuracy.


> Also, it MinutesPlayedAggregator inherits from EventDatasetAggregator, but we do some of these operations on TrackingDataset right?
> 
All operations can currently be done on an EventDataset (the information about whether a player was on the pitch or not is extracted from his positions which include a start and end time).

UnravelSports:
From my experience Opta is not great at tagging out events.  

I think it makes sense if we simply include everything that should create, or happen within a Dead Ball state simply to cover all bases. (I don't really see a downside to this).

Using receival_timestamp should be fine if it exists indeed!

UnravelSports:
A couple of this:

1. Should we add a test to assert the times are the same for both teams? Just as a sanity check.

```python
teams = list(team_minutes_played_map.keys())
assert team_minutes_played_map[teams[0]] == team_minutes_played_map[teams[1]]
```

2. I think it might be inline with the rest of kloppy's functionality to allow the breakdown key to be the string representation of the BreakdownKey too. So, "possession_state" and "position".

```python
minutes_played_per_possession_state = dataset.aggregate(
    "minutes_played", breakdown_key="possession_state"
)
```

I haven't looked at the actual inner workings.

Issue #442, opened 2025-03-27
[Sportec] Missing Events and Qualifiers


UnravelSports:
Hi,

I was going over the parsed, implemented, not supported and unknown EventTypes for Sportec for the upcoming documentation release and below is a list of all the events that are currently not implemented:

- `OtherBallAction` > `DefensiveClearance` (either "true" or "false")
- `BallClaiming` > `InterceptedBall` (I don't know if this is a Goalkeeper event or a regular event)
- `TacklingGame` > `DribbleEvaluation` (either "successful" or "unsuccessful")
- `Caution` > `CardColor` (either "yellow" or "red", and maybe "secondYellow" but that I'm not sure of)

Then we also have some Kloppy types that I don't know exist in Sportec. We might need Sportec documentation to figure this out:

- `EventType.PLAYER_ON` `EventType.PLAYER_OFF`
- `EventType.RECOVERY`
- `EventType.MISCONTROL`

Then, for PassQualifiers I found (there may be more in the docs):

- `<Pass Direction="throughBall"/>`
- `<Pass Direction="squarePass"/>`
- `<Pass Direction="diagonalBall"/>` (is this `PassQualifier.SWITCH_OF_PLAY`?)
- `<Pass Direction="backPass"/>` (this doesn't exist yet as a PassQualifier in Kloppy)

Finally, the ShotAtGoal and SavedShot items have some interesting features we are currently not processing:
- `<ShotAtGoal ChanceEvaluation="chance" xG="0.0299" CounterAttack="false" Pressure="0.79">`
- `<SavedShot SaveType="arms" SaveResult="held"/>`
     - `SaveResult="notCleared",` `SaveResult="held"`, `SaveResult="cleared"` (there can be more in the docs?)

mrkthmpsn:
I think `CardColor` is already being processed: https://github.com/mrkthmpsn/kloppy/blob/887b5ef1e1fb219ab83cf389b890d258ba4c8132/kloppy/infra/serializers/event/sportec/deserializer.py#L417 ?

Would it be good to change the file used in the event data test to one of the newer, open matches? They're much more extensive than the current file. Or would these additions be unlikely to be something to be tested? I can see in the current Sportec tests that it's mainly testing general deserialization and some specific pass-receiver logic

UnravelSports:
I might've missed that cards are already processed!

[...]

PR #480, opened 2025-06-11
[SPORTEC] Add some missing Sportec events

mrkthmpsn:
Picking up some of the things highlighted in #442 

Of those items:
- `CardColor` and `Recovery` are already being processed
- This PR adds Interception and Clearance events quite straight-forwardly
- This PR also adds most of the `TacklingGame` variations that I found in open data files. Of those that I identified, I'm only unsure about `WinnerResult=="layoff"` variations 
- I haven't touched the identified pass and shot qualifiers

The PR also adds a logic step to order events by timestamp before iterating over the events list to parse them. In the open data files, I found that some of these events were all at the end of the file, which interfered with assigning match periods (everything got put in the second half).

As mentioned in #442, this PR also adds one of the open event and metadata files as test files. There's a new deserialization test based on these files, as the old/existing test files don't include a lot of them.


probberechts:
**TODO**

- [ ] Update `docs/reference/event_data/spec.yaml`
- [ ] The implementation still seems to contain a bug where a `player_id` is passed to EventFactory instead of a `Player` object
- [ ] It would be good to add more tests for each event type. Simply checking if the output contains an instance of each event type is not sufficient

PR #499, opened 2025-09-04
feat(sportscode): add support for multiple labels per group in serializer

koenvo:
Allow serializing labels with list values to produce multiple <label> elements for the same group. Skips boolean values in lists to maintain clean output.


koenvo:
TODO: update docs https://kloppy.pysport.org/user-guide/exporting-data/sportscode/

PR #504, opened 2025-10-01
[Impect] add support for Impect event data

DriesDeprest:
Adding support for Impect as an event provider. I've used the implementation of the StatsBomb deserializer as an inspiration source.

### DIFF
--- a/kloppy/infra/serializers/event/impect/deserializer.py
+++ b/kloppy/infra/serializers/event/impect/deserializer.py
@@ -202,9 +202,9 @@ class ImpectDeserializer(EventDataDeserializer[ImpectInputs]):
                         f"Unknown position {position_key}, defaulting to Unknown"
                     )
                     position = PositionType.Unknown
-                player_starting_positions[player_starting_info["playerId"]] = (
-                    position
-                )
+                player_starting_positions[
+                    player_starting_info["playerId"]
+                ] = position
 
             players = []
             for player in team_info["players"]:


kloppy/_providers/impect.py
```diff
@@ -0,0 +1,43 @@
...
+) -> EventDataset:
+    """
+    Load Impect event data into a [`EventDataset`][kloppy.domain.models.event.EventDataset]
+
+    Parameters:
+        event_data: filename of json containing the events
+        lineup_data: filename of json containing the lineup information
+        event_types:
+        coordinates:
+        event_factory:
```
probberechts, line 26: Args:
        event_data: JSON feed with the raw event data of a game.
        lineup_data: JSON feed with the corresponding lineup information of the game.
        event_types: A list of event types to load.
        coordinates: The coordinate system to use.
        event_factory: A custom event factory.

    Returns:
        The parsed event data.

kloppy/_providers/impect.py
```diff
@@ -0,0 +1,43 @@
...
+    """
+    deserializer = ImpectDeserializer(
+        event_types=event_types,
+        coordinate_system=coordinates,
+        event_factory=event_factory
+        or get_config("event_factory")
+        or ImpectEventFactory(),
```
probberechts, line 33: This `ImpectEventFactory()` is copy-pasted from the StatsBomb example but it is doesn't have any function. I think it can also be removed from the StatsBomb code by the way.

kloppy/domain/models/common.py
```diff
@@ -1477,6 +1500,14 @@ def value(self) -> Optional[float]:
...
 
 
+@dataclass
+class PressingIntensity(ScalarStatistic):
+    """Pressing Intensity"""
+
+    def __post_init__(self):
+        self.name = "Pressing Intensity"
+
+
```
probberechts, line 1510: I don't think this belongs in this commit.

kloppy/domain/models/pitch.py
```diff
@@ -699,3 +699,24 @@ class WyscoutPitchDimensions(PitchDimensions):
     corner_radius: float = 0.97  # inferred
     penalty_spot_distance: float = 10.0
     penalty_arc_radius: float = 7.74  # inferred
+
+
+@dataclass
+class ImpectPitchDimensions(PitchDimensions):
```
probberechts, line 705: These are `MetricPitchDimensions`. It is not needed to create a separate object.

probberechts (commented):
Did a quick first scan of the code. In general, this looks good. Just a few minor things for now. For the docs, would it be possible to update the `docs/reference/event-data/spec.yaml` file too?

UnravelSports:
Updated TODO for this PR:
- [X] @koenvo adds Team and Player file support
- [X] @koenvo will check against video 
- [ ] @koenvo will assign players that come of the field with no sub as the PlayerOff event type
- [X] @koenvo will add `load_open_data`, related:
     - [ ] @Flosch1006 we will need the new repo to be open to make sure the new URL is correct
     - [ ] @koenvo change default `match_id="100214"` to `match_id="188164"` the former Id is not in the open dataset
 - [ ] @koenvo change test file from `impect_meta.json` to `impect_lineups.json` to fix failing test
 - [ ] @koenvo will check v4 (add as tests)
 - [ ] @koenvo check why `CardType` is not showing in `to_df`
 - [ ] @koenvo check `is_under_pressure`. We're not seeing expected `PressureEvents`

@DriesDeprest if you have some time, perhaps you can look at the bottom 2 points too.

Flosch1006:
here is the link to the open data repository: https://github.com/ImpectAPI/open-data

UnravelSports:
> here is the link to the open data repository: https://github.com/ImpectAPI/open-data

@Flosch1006 but you were going to put it on a clean repo right? Or did I misunderstand?

Flosch1006:
> @DriesDeprest 
> 
> I get the following error when trying to open 122838 with `impect.load_open_data(match_id="122838")`:
> 
>  ```python
> File ~/PycharmProjects/kloppy/kloppy/infra/serializers/event/impect/specification.py:299, in SHOT._create_events(self, event_factory, teams, **generic_event_kwargs)
>     296 shot_dict = self.raw_event["shot"]
>     297 result = self.RESULT(self.raw_event["result"])
> --> 299 shot_end_coordinates, shot_result = parse_shot_end_coordinates(
>     300     shot_dict, result
>     301 )
>     303 body_part = BODYPART(self.raw_event["bodyPartExtended"])
>     305 qualifiers = _get_body_part_qualifiers(body_part)
> 
> File ~/PycharmProjects/kloppy/kloppy/infra/serializers/event/impect/helpers.py:210, in parse_shot_end_coordinates(shot_info, shot_result)
>     207     result = ShotResult.GOAL
>     208 # elif is_own_goal:
>     209 #     result = ShotResult.OWN_GOAL
> --> 210 elif shot_info["woodwork"]:
>     211     result = ShotResult.POST
>     212 elif abs(y_coordinate) < 3.66 and z_coordinate < 2.44:
> 
> KeyError: 'woodwork'
> ```
> 

'woodwork' is a key that was added in data version V4. Therefore we removed it from the open data as it would always be null (open data is V3). Sorry for causing hiccups here.

I will send you a list of keys that were removed on Monday morning.

Flosch1006:
Here is the promised list of removed keys:

- `dribble`
- `opponent`
- `setpiece`
- `shot.gk`
- `shot.woodwork`

UnravelSports:
Awesome, so we have a couple more things remaining now. I've added one thing unrelated to the PR for myself here too, so I can have everything in one place.

- [ ] @koenvo will assign players that come of the field with no sub as the PlayerOff event type
 - [ ] @koenvo will check v4 (add as tests)
 - [ ] @koenvo check why `CardType` is not showing in `to_df`
 - [ ] @koenvo check `is_under_pressure`. We're not seeing expected `PressureEvents`
 - [ ] @UnravelSports share basic how-to load Impect data with Kloppy for in the open-data repo readme with @Flosch1006

koenvo:
Right now we only have a PressureEvent which means a player is pressuring. There is still an ongoing discussion on how to handle the case where events have a "pressure score". Can you remember @probberechts ?

probberechts:
> Right now we only have a PressureEvent which means a player is pressuring. There is still an ongoing discussion on how to handle the case where events have a "pressure score". Can you remember @probberechts ?

See here: https://github.com/PySport/kloppy/pull/296#issuecomment-2550850683

UnravelSports:
One final thing, IMPECT has a license similar to StatsBomb. When we load StatsBomb open data we get a warning about this license. Perhaps we should add this to the impect open data loader too.

```python
warnings.warn(
        "\n\nYou are about to use IMPECT public data."
        "\nBy using this data, you are agreeing to the user agreement. "
        "\nThe user agreement can be found here: https://github.com/ImpectAPI/open-data/blob/main/LICENSE.pdf"
        "\n"
    )
```

PR #510, opened 2025-10-22
Auto resolve GitHub urls

koenvo:


koenvo:
> I've never looked at this in detail, but I think you can also implement this with an FSSpecAdapter: https://filesystem-spec.readthedocs.io/en/latest/_modules/fsspec/implementations/github.html

It looks good. Only I'm a bit afraid it will introduce quite some overhead. It will execute `ls` for each file. Under the hood it seems to use `HTTPFileSystem`.

